PR #101, opened 2015-05-15
5.6.3 tk

twkrol:
Added message counter to mailbox General page.
(you may skip the vcxproj changes commit [ffb32ca87395a214966e8b69daa297b9ac1b4c1d] that adapts new versions of OpenSSL and Boost)

It's my first pull request ever - please be aware of newbie ;-)


martinknafve:
This won't be included. The reason this isn't already there is because this change will load all the account's messages from the database, which is potentially really slow.

PR #193, opened 2016-08-04
Include HTMLBody into IMAP TEXT search.

M0ns1gn0r:
In contrast to the `SEARCH BODY`, the `SEARCH TEXT` command didn't perform search in HTML bodies of messages . One had to resort to `SEARCH OR SUBJECT "abc" BODY "abc"`.


ericnewton76:
Wow, does hmailserver not have a better way to search, besides loading every single message, one by one, and searching within the message body?

Seems like the search should get deferred into the underlying data store.

PR #202, opened 2016-10-29
Update hMailServerInnoExtension.iss - new switch

bholliger:
Hi there!
Before the MS SQL CE implementation, the setup with silent switch installed no internal database. In order to have the same behaviour of the setup, I've added the 'useinternaldbms' switch.
Starting setup with the switch /useinternaldbms:false will set the radio from default "use internal database" to "use external database" and during a silent install, it leaves the setup without a database management system.
I'd love to see this change in future versions of hmailserver. Thanks for pulling this change.
Have a great day!
-Bruno.


martinknafve:
Thanks!

PR #204, opened 2016-11-22
Fixed implicit conversion: "int" to "unsigned char"

EpicCoder:


jwdonahue:
Where the values are going to be assigned to boost::asio::ip::address_v4::bytes_type variables, why not use that instead of unsigned char?

PR #210, opened 2017-01-13
Add moving rules and rule actions up and down in webadmin

tunis78:
Added function in PHPWebAdmin to move rules and rule actions up and down to change order.


martinknafve:
Sorry, not merging this one. I know it's a very old PR, but I won't merge any substantial changes to web admin due to lack of time.

PR #224, opened 2017-05-18
Update DNSResolver.cpp to prefer IPv6

Claus-Mtsn:
If IPv6 is available, this should be put in the result as the primary record. IPv4 is fallback.

Today IPv4 is prefered from a subdomainname containing both A & AAAA records. This should be changed to fully support IPv6.

### DIFF
--- a/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp
+++ b/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp
@@ -243,11 +243,11 @@ namespace HM
          return false;
       }
 
-      bool ipv4AddressResult = Resolve_(sDomain, vecFoundNames, DNS_TYPE_A, 0);
-
       bool ipv6AddressResult = Configuration::Instance()->IsIPv6Available() ?
                                  Resolve_(sDomain, vecFoundNames, DNS_TYPE_AAAA, 0) :
                                  false;
+	   
+      bool ipv4AddressResult = Resolve_(sDomain, vecFoundNames, DNS_TYPE_A, 0);
 
       return ipv4AddressResult || ipv6AddressResult;
    }


gicldk:
The reason for this change can be found here.

https://www.hmailserver.com/forum/viewtopic.php?f=7&t=31259

OutbackMatt:
I have IPv6 available on my LAN, and especially on my mail server.
My ISP doesn't offer IPv6 at the moment.

if IPv6 was preferred, then no mail would leave my LAN, even thought IPv6 addresses are found by the DNS etc. If this were to be implemented, there would need to be some form of switch in hmailserver so that we could turn it off and use hMailserver with current functionality.

Also to achieve what you are trying to do, doesn't the order of the returns (line 252) need to be swapped around to make IPv6 the first address returned. Haven't you only swapped the order of evaluation, not the order of return.

gicldk:
> I have IPv6 available on my LAN, and especially on my mail server. My ISP doesn't offer IPv6 at the moment.

This is not a problem. You've probably enabled IPv6 on your LAN to gain knowledge on the subject. hMailServer is dualstack capable, and as such, your network not being connected to the internet on IPv6.

My ISP doesn't either support IPv6, but I've gained IPv6 capabilities anyway by implementing the HE Tunnelbroker. You could as well and it's free of use.

> if IPv6 was preferred, then no mail would leave my LAN, even thought IPv6 addresses are found by the DNS etc.

Actually this is not correct. hMailServer will try all mx A & AAAA-records several times before giving up. In your case, where IPv6 is preferred, hMailServer will try to connect to the IPv6 address of ie. gmail.coms 1st. mx and will time-out. Afterwards it will connect to the IPv4 address and deliver the email.

You can make your own test by making entries into the hostfile in C:\Windows\System32\drivers\etc. Personally I use a local ISP, which has an dualstackmailserver; mail.gigabit.dk. To test, I've put these records in:
	2a00:7660:0:24::88	mail.gigabit.dk
	172.27.131.10		mail.gigabit.dk #Bogus ip, where no SMTP conn is avaible.
	185.24.168.136		mail.gigabit.dk

in this case hMailServer resolves the IPv4 first, but identifies all 3 records. Fails to deliver to the bogus but delivers to the valid record. Here is an output from the log:
"TCPIP"	2788	"2017-05-19 09:57:28.493"	"DNS MX lookup: gigabit.dk"
"TCPIP"	2788	"2017-05-19 09:57:28.499"	"DNS - MX Result: 3 IP addresses were found."
"DEBUG"	2788	"2017-05-19 09:57:28.515"	"Starting external delivery process. Server: mail.gigabit.dk (172.27.131.10), Port: 25, Security: 2, User name: "
"DEBUG"	2788	"2017-05-19 09:57:28.531"	"Creating session 2914"
"TCPIP"	2788	"2017-05-19 09:57:28.546"	"Connecting to 172.27.131.10:25..."
"DEBUG"	4900	"2017-05-19 09:57:49.597"	"SMTPDeliverer - Message 59217 - Connection failed: Host name: 172.27.131.10, message: The semaphore timeout period has expired"
"DEBUG"	4900	"2017-05-19 09:57:49.597"	"Ending session 2914"
"DEBUG"	2788	"2017-05-19 09:57:49.628"	"External delivery process completed"
"DEBUG"	2788	"2017-05-19 09:57:49.644"	"Starting external delivery process. Server: mail.gigabit.dk (185.24.168.136), Port: 25, Security: 2, User name: "
"DEBUG"	2788	"2017-05-19 09:57:49.644"	"Creating session 2917"
"TCPIP"	2788	"2017-05-19 09:57:49.659"	"Connecting to 185.24.168.136:25..."
"DEBUG"	4900	"2017-05-19 09:57:49.691"	"TCP connection started for session 2917"
"DEBUG"	1496	"2017-05-19 09:57:49.737"	"Performing SSL/TLS handshake for session 2917. Verify certificate: True, Expected remote host name: mail.gigabit.dk"
"DEBUG"	1672	"2017-05-19 09:57:49.784"	"Certificate verification succeeded for session 2917."
"TCPIP"	1496	"2017-05-19 09:57:50.331"	"TCPConnection - TLS/SSL handshake completed. Session Id: 2917, Remote IP: 185.24.168.136, Version: TLSv1.2, Cipher: DHE-RSA-AES256-GCM-SHA384, Bits: 256"
"DEBUG"	1188	"2017-05-19 09:57:50.394"	"Ending session 2917"
"DEBUG"	2788	"2017-05-19 09:57:50.409"	"External delivery process completed"

> If this were to be implemented, there would need to be some form of switch in hmailserver so that we could turn it off and use hMailserver with current functionality.

I can't say if this is very usefull. I gather that normal, you wouldn't configure IPv6, the way you have done it. All users are getting a /64 addressscope which gives you 18 trillions public addresses. You don't use NAT, but routes trafic instead. You still use firewall to block unwanted traffic like you do today with portforwarding. In your case, the only downside will be the extra delivery attempt, that results in a timeout.

OutbackMatt:
If hMailserver tries to send to gmail.com, hMailserver gets a long list of possible mx addresses, and tries to deliver first to one, then the next.

We have a 'hidden' setting in the hmailserver.ini that details how many MX servers to try. This is intended to be useful when sending to domains that greylist.

MXTriesFactor=3
; This key defines a multiplier to determine how many MX/A server results to try per retry.
; For example: If set to 2 & the recipient has 11 servers, on 1st try, 1 x MXTriesFactor or 2 are tried.
; On 2nd try, 2 x MXTriesFactor or 4 are tried. On 3rd 6 are tried & so on. Eventually all 11
; would be tried if the message was not successively delivered enough times.
; This defaults to disabled if not defined


gMail has multiple IPv6 MX records too. I wouldn't want to circle through them all before an IPv4 was used.

ALSO, my set up is standard windows default, except that I have specifically enabled IPv6 on my router for my LAN. Windows machines by default have IPv6 enabled (and often when there is networking issues I turn IPv6 off and intermittent windows connection issues just disappear.)

gicldk:
> gMail has multiple IPv6 MX records too. I wouldn't want to circle through them all before an IPv4 was used.

If you resolve gmail.com it has these mx's as you describe:
5 gmail-smtp-in.l.google.com.
10 alt1.gmail-smtp-in.l.google.com.
20 alt2.gmail-smtp-in.l.google.com.
30 alt3.gmail-smtp-in.l.google.com.
40 alt4.gmail-smtp-in.l.google.com.


As far as I can tell from the source, this leads to a sortet list of IPs:
5 209.85.233.26
5 2a00:1450:4010:c08::1b
10 108.177.97.26
10 2404:6800:4008:c00::1a

...and so on.

Lets take your setup as a reference.

First it wil attemp to deliver on 209.85.233.26:25. For the sake of the example, lets just assume it is down. After 20 sek. it times-out.
Your server will now try the next in line and attempt to deliver on 2a00:1450:4010:c08::1b:25. This will fail as well, because you don't have IPv6 capable internet access.
Your server will now try the next in line and attempt to deliver on 108.177.97.26:25. This server answers the request and your email is delivered.

You can try this your self, by putting the first mx in your hostfile, with a spoofed local IPv4 address, that doesn't smtp/25.

What my change will do, is the making the IP list look like this:
5 2a00:1450:4010:c08::1b
5 209.85.233.26
10 2404:6800:4008:c00::1a
10 108.177.97.26

> ALSO, my set up is standard windows default, except that I have specifically enabled IPv6 on my router for my LAN. Windows machines by default have IPv6 enabled (and often when there is networking issues I turn IPv6 off and intermittent windows connection issues just disappear.)

I don't see how a misconfiguration has to do with the features of software. If you wan't to run software IPv4 only, then disable the IPv6 network. Alternativly I don't see no problem in moving the IPv6 configuration enabling feature from an automecanism to a configuration item in the software.

martinknafve:
Wouldn't this slow down outbound email for a huge amount of hMailServer users? I think IPv6 is enabled by default in Windows, so hMailServer would typically prefer that. But at the same time relatively few ISP's support IPv6 (basically "default" is IPv4 only). I'm not sure what benefit this change would give?

mstenz:
https://www.internetsociety.org/resources/2018/state-of-ipv6-deployment-2018/
This is from 2018. I would assume today Ipv6 is relatively widely used/available. 
You won't believe how much faster internet is with IPv6 enabled. The reason for this is that ipv6 enabled pop's use the latest hardware and most likely higher bandwidth. Providers overcome historical grown infrastructure with IPv6 and this resulted with a lower latency as well. 
i definitive prefer using IPv6 if possible. Maybe it would be a configuration setting so everyone decide on their own?

mstenz:
Here an example of routing-differences between Ipv4 and IPv6 from same source to same target. 

> C:\WINDOWS\system32>tracert mail.cowmedia.de
> 
> Routenverfolgung zu mail.cowmedia.de [2a02:c205:1000:6043::3]
> über maximal 30 Hops:
> 
>   1    <1 ms    <1 ms    <1 ms  fritz.box [2003:f0:e3c9:6800:464e:6dff:fea3:d985
> ]
>   2     5 ms     4 ms     4 ms  2003:0:8805:9800::1
>   3     7 ms     6 ms     6 ms  2003:0:1804:4000::1
>   4     8 ms     8 ms     8 ms  2003:0:1804:4002::2
>   5    24 ms     7 ms     7 ms  2001:a60::89:303
>   6    70 ms    44 ms    13 ms  2001:a60:902c::2
>   7    18 ms    21 ms    17 ms  2a02:c205:1000:6043::3
> 
> Ablaufverfolgung beendet.
> 
> C:\WINDOWS\system32>tracert -4 mail.cowmedia.de
> 
> Routenverfolgung zu mail.cowmedia.de [91.205.172.53]
> über maximal 30 Hops:
> 
>   1     1 ms    <1 ms    <1 ms  fritz.box [192.168.2.1]
>   2     6 ms     5 ms     5 ms  62.155.242.178
>   3    14 ms    14 ms    14 ms  217.5.116.230
>   4    14 ms    13 ms    13 ms  217.5.116.230
>   5     *       20 ms    20 ms  ffm-b4-link.telia.net [213.248.93.186]
>   6    63 ms    65 ms    66 ms  ffm-bb3-link.telia.net [62.115.114.88]
>   7    22 ms    21 ms    21 ms  mcn-b2-link.telia.net [62.115.120.132]
>   8    20 ms    19 ms    19 ms  gw02.contabo.net [213.248.101.78]
>   9    20 ms    22 ms    19 ms  mail.cowmedia.de [91.205.172.53]
> 
> Ablaufverfolgung beendet.

martinknafve:
This won't be merged unless someone can show some stats showing that a majority of all email servers listen on IPv6.

Dravion:
A Windows Admin can adjust which DNS-Resolver should be preferred with:

#Prefer IPv4 over IPv6
reg add hklm\system\currentcontrolset\services\tcpip6\parameters /v DisabledComponents /t REG_DWORD /d 0x20 

#Prefer IPv6 over IPv4
reg add hklm\system\currentcontrolset\services\tcpip6\parameters /v DisabledComponents /t REG_DWORD /d 0x0

mstenz:
@Dravion hmailserver should if possible not use any OS settings because environments can be compley and maybe the OS should use IPv6 for standard dns while hmailserver should be other prefer IPv4 vs IPv6. This really depends on the environment and the connection to other (internal) systems). 
I still prefer the option that the hmail administrator can configure the expected behaviour for his instance directly in the configuration - thats what a professional software do.

RvdHout:
My ISP recently enabled IPv6, been playing with it for a week now.
Trying to test my IPv6 SPF records i encountered the same issue as the others described here and i was unable to send out using IPv6.
The DNSResolver::GetIpAddresses() function always inserts the found IPv4 addresses above the found IPv6 addresses which basically forces hMailServer to try delivery over IPv4 first.

For the fun of it to test/verify this behavior i inserted a shuffle inside DNSResolver::GetIpAddresses() function
`td::random_shuffle(foundValues.begin(), foundValues.end()); `
and you guessed all of the sudden hMailServer started deliveries over IPv6 (completely random) but this enabled me to test my SPF IPv6 record (with macro expanding)

RvdHout:
@martinknafve 

I don't think this necessarily slows down hMailServer in a IPv4 only environment as IPv6 addresses are only queried when:Configuration::Instance()->IsIPv6Available()

Is it OK if i give a go at it? Preferably add it as a optional setting under SMTP (deliveries)?

RvdHout:
PreferIPv6

Something like this?

PR #238, opened 2017-10-26
Implemented per-user flags for public folders

94ivana:
Added funcionality for per-user flags using aditional table in the database that keeps track which user read which message in a public folder.


martinknafve:
This won't be merge due to lack of time.

Neustradamus:
@94ivana: Can you look to add a new PR with your changes?

Thanks in advance.

Issue #250, opened 2018-02-27
Fetch account deliver to MIME recipients logic


sprintman:
Sometimes I get some a message, addressed to two recipients where the sending server has decided to send to each recipient separately. Thus in the external account I end up with two messages, each with one recipient listed in the Envelope-To header, but both recipients listed in the To header. When hMailServer fetches the messages, it delivers both to both recipients.

When it exists, the (X-)Envelope-To/X-RCPT-TO header (when it exists) explicitly tells us who the message was sent to and we should not use the To/CC/Received headers.

I have implemented the following changes to the external fetch:
First add all recipients from X-RCPT-TO, X-Envelope-To and Envelope-To header. Only if there are no valid recipients, add all recipients from To, CC and Received headers.

Note, that I have added Envelope-To as an additional header to check. There are a few more that probably should be added, but I was not sure if they contain the SMTP recipient or the final recipient after following aliases etc. Ultimately it would be good to have an option for which headers to check as it seems there are many in use.

martinknafve:
Since there are pull requests for this I'll close this issue.

PR #251, opened 2018-02-27
External Fetch: First try (X-)Envelope-To and X-RCPT-TO headers, then…

sprintman:
#250 … fallback to TO, CC and Received


martinknafve:
We will merge https://github.com/hmailserver/hmailserver/pull/428 instead. It solves the same issue, but in a more flexible way.

Issue #287, opened 2019-07-05
Ip Range Allow deliveries/Require SMTP authentication Bug


RvdHout:
Have been fiddling around with the IP ranges, when I disallow Deliveries from External to External for a IP Range but do not uncheck External to External in Require SMTP authentication the error message displayed is:
"530 SMTP authentication is required."

Only if I also uncheck External to External in Require SMTP authentication the error message is correct. eg:
"550 Delivery is not allowed to this address."

This seems incorrect to me...Don't you think?
I think settings in Allow deliveries must always be the preferred and appointed criteria above settings in Require SMTP authentication, in Administrator GUI this looks to be the case as the Require SMTP authentication checkbox is greyed-out/readonly as soon you uncheck Allow deliveries, but in the code it seems to function visa versa

https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/SMTP/SMTPConnection.cpp#L681

mstenz:
I have experienced same issue also after first configuration of the server, it was quite confusing.

RvdHout:
I think simply flipping over the order of execution fixes this...weird i never noticed this before

RvdHout:
https://www.hmailserver.com/forum/viewtopic.php?f=8&p=213151#p213151

martinknafve:
@RvdHout I fixed this before I noticed your pull request. I'll compare our fixes and see if they are the same.

martinknafve:
Merged

Issue #267, opened 2018-08-21
SMTP multiply max message size with 1024


Virinum:
I've set the max message size to 102400 KB (100 MB). Via telnet I receive `250-SIZE 102400000`. But 102400000 Bytes are 97.65625 MB. So if I want exactly 100 MB, I've to set the value to 104857.6 (but points aren't allow).

E.g. Gmail returns `250-SIZE 157286400` (150 MB). It's not possible for me to set that value, too.

So here my suggestion: Multiply the max message value with 1024 (not 1000).

RvdHout:
SMTPConection.cpp

**old**
`      // Append size keyword
      {
         String sSizeKeyword;
         int iMaxSize = smtpconf_->GetMaxMessageSize() * 1000;
         if (iMaxSize > 0)
            sSizeKeyword.Format(_T("\r\n250-SIZE %d"), iMaxSize);
         else
            sSizeKeyword.Format(_T("\r\n250-SIZE"));
         sData += sSizeKeyword;
      }`

**new**
`      // Append size keyword
      {
         String sSizeKeyword;
         int iMaxSize = smtpconf_->GetMaxMessageSize() * 1024;
         if (iMaxSize > 0)
            sSizeKeyword.Format(_T("\r\n250-SIZE %d"), iMaxSize);
         else
            sSizeKeyword.Format(_T("\r\n250-SIZE"));
         sData += sSizeKeyword;
      }`

martinknafve:
Included in 5.7.

Issue #265, opened 2018-08-11
C++ Compiler - Autoembedded Spy/Telemetry code


Dravion:
Right now its only reported for Visual C/C+++ 2015 but i heard something similar on 2013 as well
https://stackoverflow.com/questions/37761768/how-to-prevent-visual-studio-2015-update-2-to-add-telemetry-main-invoke-trigger
![lqbls](https://user-images.githubusercontent.com/15117429/43991254-13ca59a0-9d69-11e8-8868-7dcb1794595d.jpg)

Issue #172, opened 2016-05-14
DKIM on rule 'reply' not applied


jimimaseye:
Scenario:

1, DKIM signing setup and enabled - Sending an email with email client signs the email correctly (with the usual "dkim-signature:" header).

2, Set up an Account Rule with a "REPLY (email)" as an action (all fields filled correctly).

BUT the received reply email (sent by the rule) does not get DKIM signed.

Full details and headers showing the problem here: https://www.hmailserver.com/forum/viewtopic.php?f=7&t=29799

RvdHout:
https://www.hmailserver.com/forum/viewtopic.php?f=7&t=33976&p=211807#p211807

martinknafve:
Included in 5.7.

PR #294, opened 2019-08-16
https://www.hmailserver.com/forum/viewtopic.php?f=10&p=214033#p214033

RvdHout:
SMTP multiply max message size with 1024 issue #267
Add email address variable to SignatureAdder.cpp pull #265
DKIM on acccount-rule 'reply' not applied #172 issue #172
preserve RewriteEnvelopeFromWhenForwarding setting when forwarding from account rule
The logical flow should be to disregard "Require SMTP authentication" if "Allow deliveries from" is unselected issue #287


martinknafve:
I'll merge this, but from now on please limit the commits and pull requests to 1 fixed issue in each. When there are multiple issues fixed in one commit/issue, I can't select what parts to merge in some easy way (I think).  Also, looking through the git history will be more confusing if there are unrelated changes made in the same commit.

RvdHout:
> I'll merge this, but from now on please limit the commits and pull requests to 1 fixed issue in each. When there are multiple issues fixed in one commit/issue, I can't select what parts to merge in some easy way (I think). Also, looking through the git history will be more confusing if there are unrelated changes made in the same commit.

Sorry, will do that next time

martinknafve:
No prob, thanks for the fixes. :)

PR #297, opened 2019-08-18
Update IMAPCommandLogin.cpp

RvdHout:
Think this is better, otherwise the OnClientLogon EventHandler will not be triggered properly when non-full emailaddress is used and no default domain exists


RvdHout:
Basically this is the same behavior as how it currently works in the 5.6.8 branch....
Not sure why you changed it for 5.7, as i don't see the actually use for it 

https://github.com/hmailserver/hmailserver/blob/5.6.8/hmailserver/source/Server/IMAP/IMAPCommandLogin.cpp

PR #301, opened 2019-08-29
DKIM sign NDR messages (forwarded to external)

RvdHout:
This change adds the possiblity/ability to DKIM (and DMARC) sign NDR messages forwarded to external-accounts.
Even if RewriteEnvelopeFromWhenForwardingsetting is set NDR messages EnvelopeFrom will not be rewritten to localaccount address but uses mailer-daemon@ instead as EnvelopeFrom address.

**Requirements to DKIM sign NDR messages**
If the mailer-daemon@ domain exists on the server and DKIM is enabled for that domain the NDR message forwarded to external domains will be DKIM (and optionaly DMARC) signed

### DIFF
--- a/hmailserver/source/Server/Common/Application/IniFileSettings.cpp
+++ b/hmailserver/source/Server/Common/Application/IniFileSettings.cpp
@@ -139,6 +139,8 @@ namespace HM
 
       max_no_of_external_fetch_threads_ = ReadIniSettingInteger_("Settings", "MaxNumberOfExternalFetchThreads", 15);
       add_xauth_user_header_ = ReadIniSettingInteger_("Settings", "AddXAuthUserHeader", 0) == 1;
+
+      daemonaddress_domain_ = ReadIniSettingString_("Settings", "DaemonAddressDomain", "");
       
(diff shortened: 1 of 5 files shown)


mstenz:
Can you please ellaborate more on the "mailer-daemon@" domain? do you really mean a domain or only the account/alias on the different domains? in the documentation there seems to be nothing really about this

RvdHout:
https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/Util/MailerDaemonAddressDeterminer.cpp

mstenz:
i am no developer, so cannot read the code.

RvdHout:
> i am no developer, so cannot read the code.

That is why i wrote a description underneath it!

mstenz:
thx. i see your update now

RvdHout:
Last extra commit (INI setting) is to overrule 'Host name' set in SMTP settings, which in real world scenarios would be the PTR DNS records of you IP address. This can be usefull in cases where, for example if your PTR is assigned by your ISP and you have no control over it.  Or you simply like another domain to be used as NDR address

You can set the 'DaemonAddressDomain' in the INI under [Settings]

```
[Settings]
DaemonAddressDomain=example.com
```

PR #312, opened 2020-01-29
Update SURBL.cpp

RvdHout:
Better URL regex for SURBL?

For more info, checkout this topic: https://www.hmailserver.com/forum/viewtopic.php?f=7&t=34743

### DIFF
--- a/hmailserver/source/Server/Common/AntiSpam/SURBL.cpp
+++ b/hmailserver/source/Server/Common/AntiSpam/SURBL.cpp
@@ -38,9 +38,9 @@ namespace HM
       String sBody = pMessageData->GetBody() + pMessageData->GetHTMLBody(); 
 
       // Extract URL's from the mail body:
-      // Original: https?:\/\/([^?><\\ \"'\/]*)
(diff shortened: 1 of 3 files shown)


martinknafve:
It looks like this pull request contains a mix of various things now. Each individual change looks OK, but I'm not sure if they were added on purpose to this pull request?

RvdHout:
I didn't add those on purpose to this pull request , isn't this done automatically when i commit something else to my fork?

RvdHout:
M....i see, I think i have to:

1. Create a branch, commit & create pull request
2. Then delete that branch, and repeat step 1 for every commit i make, right?

mstenz:
Yes, deleting a branch ist best practice.

martinknafve:
Yes, I think you're right. I'm actually not 100% sure myself. If it's not too much work, could you try it for the SURBL change and see if that works it?

RvdHout:
Mmmm, looks like it closes the pull request immediately after i delete the branch... doesn't seem the  proper way of doing this

PR #319, opened 2020-03-07
Update DNSResolver.cpp

RvdHout:
DKIM signature verification when published DNS record is CNAME

Only seems to be a issue on 5.7.x

For details see:
https://www.hmailserver.com/forum/viewtopic.php?f=7&t=34834&p=218884#p218884

### DIFF
--- a/hmailserver/source/Server/Common/Application/Errors.txt
+++ b/hmailserver/source/Server/Common/Application/Errors.txt
@@ -1,4 +1,5 @@
-4401, Too many recursions during IP adress lookup. Query: {0}
+4401, Too many recursions during IP address lookup. Query: {0}
+4402, Too many recursions during TXT record lookup. Query: {0}
 5121, PersistentACLPermission::SaveObject, ACL item of type PTAnyone not initialized correctly.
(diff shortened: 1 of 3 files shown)


martinknafve:
I think this change introduces a potential stack overflow in the TXT lookup. If you do a TXT lookup for a host name which only has a CNAME pointing at itself, then I think the function GetTXTRecords would just call itself in a loop until it crashes.   The same would happen if host a.example.com has a CNAME for b.example.com and b.example.com has a cname for a.example.com. 

There needs to be a limit so the function only follows say ~10 CNAME before giving up. See the method GetIpAddressesRecursive_ for example how this is done previously.

RvdHout:
Good one....
mmmm, let me think...

I don't think we need a bool followCnameRecords as used in GetIpAddresses, and always try to use CNAME when none TXT  records exists, i seen CNAME records for SPF TXT records as well (em.spotify.com) so the only addiotion should be some recursionLevel

RvdHout:
Better like this?

martinknafve:
Thanks!

PR #322, opened 2020-03-10
Update SPF.cpp

RvdHout:
Add RMSPF library SpamTestSPF Result to DEBUG logging, see https://www.hmailserver.com/forum/viewtopic.php?p=218710#p218710

### DIFF
--- a/hmailserver/source/Server/SMTP/SPF/SPF.cpp
+++ b/hmailserver/source/Server/SMTP/SPF/SPF.cpp
@@ -24,6 +24,29 @@ namespace HM
 
    }
 
+   inline const char* EnumToString(int v)
+   {
+      switch (v)
+      {
+         case SPF_Pass: 
+            return "Pass";
+         case SPF_SoftFail: 
+            return "SoftFail";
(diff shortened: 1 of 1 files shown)


RvdHout:
Mmm, after looking through the SPF code, it seems it should return some SPF test result, but it doesn't, see here: 
https://github.com/hmailserver/hmailserver/blob/7446e1c84b2391102d464a89a73a73b5387a98d3/hmailserver/source/Server/SMTP/SPF/SPF.cpp#L71 and here: https://github.com/hmailserver/hmailserver/blob/7446e1c84b2391102d464a89a73a73b5387a98d3/hmailserver/source/Server/Common/AntiSpam/SpamTestSPF.cpp#L57

I never ever have seen the SPF test any result, it just displays as "Rejected by SPF ()"

RvdHout:
I will close this pull request for now, maybe you could look into the original code and find out why it nevers returns anything for the Explanation provided by RMSPF library.... 

If that is a no go, maybe you could incorporate my workaround....

PR #324, opened 2020-03-13
Update SpamTestHeloHost.cpp

RvdHout:
SpamTestHeloHost, on localhost the reported host is the computer name, eg, it will always fail this test when send thru 127.0.0.1 and therefor should be ignored


RvdHout:
Additional info about this and the SPF check for 127.0.0.1 
https://www.hmailserver.com/forum/viewtopic.php?f=10&t=34865

RvdHout:
@martinknafve you don't agree that messages send true a local IP Address should never be checked/scored as such message never can pass those (HELO and SPF) test?

Issue #295, opened 2019-08-17
Bundle libmysql.dll with installation to make installation smoother


martinknafve:
With hMailServer 5.7, any user of MySQL will need to add place a new libmysql.dll in the bin folder, since hMailServer is now 64-bit. This will likely cause issues for a lot of users.

To remove this friction, the installer should just bundle the libmysql.dll-file. The reason it was removed years ago was because hMailServer was closed source and couldn't bundle it due to licensing, which is no longer an issue.

martinknafve:
With hMailServer 5.7, any user of MySQL will need to add place a new libmysql.dll in the bin folder, since hMailServer is now 64-bit. This will likely cause issues for a lot of users.

To remove this friction, the installer should just bundle the libmysql.dll-file. The reason it was removed years ago was because hMailServer was closed source and couldn't bundle it due to licensing, which is no longer an issue.

helptiger:
Nice, so hMailServer works with MariaDB out of the box!

OutbackMatt:
I had to upgrade my MySQL to a newer version, and then manually track down the related version of the libmysql.dll, and then I had to run the DBUpdater.exe

hMailserver prompted me correctly through the process but including the dll is a great idea.

Is there only one version of the dll?
MySQL made it seem that there are many...


@helptiger
Is MariaDB 64 bit?

Dravion:
> Nice, so hMailServer works with MariaDB out of the box!

AS long as you stay with MySQL Connector/C version 5.7.x series you can run hMailServer 5.x with MySQL. There is a MariaDB Install Guide in the hMailServer forum https://www.hmailserver.com/forum/viewtopic.php?f=21&t=32869

PS: MariaDB is not officially supported and tested, but it should work.

jimimaseye:
@martinknafve Some concerns from dravion about what dll's work and what doesn't.  Do these concerns apply?: https://www.hmailserver.com/forum/viewtopic.php?t=33531

Dravion:
Yeah, this can be a Problem.
While PostgreSQL made it very easy to compile hMailServer with latest PostgreSQL Support, simply by adding (for example: C:\Program Files\PostgreSQL\10\include\libpq-fe.h and its lib
C:\Program Files\PostgreSQL\10\lib\libpq.lib) from retail PostgreSQL, it's on the other hand not that easy to update to latest MySQL 8+ Include Headers and binary libs. I believe it has something todo,
[...]

PR #336, opened 2020-05-08
Installer bug fixes (Crt/UCrt VC++2019 Runtime and missing DLL's) 

Dravion:
This Patch contains some bug fixes within the Innosetup Installer to met the 
the new Visual Studio 2019 VC 142 and Universal C-Runtime dependencies.
 
Fixed: api-ms-win-crt-runtime-l1-1-0.dll is missing on Windows 7+SP1 / 2008 R2+SP1 x64
Fixed: api-ms-win-crt-runtime-l1-1-0.dll is missing on Windows 8 / 2012 x64
Fixed: api-ms-win-crt-runtime-l1-1-0.dll is missing on Windows 8.1 /2012 R2 x64
Fixed: libmysql.dll x64 v5.7.30 +Visual C/C++ 2013 /  Visual 2019 Crt/UCrt mismatch 
See: Issue debate: https://github.com/hmailserver/hmailserver/issues/295

Detailed change documentation can be found in:
*hmailserver\hmailserver\installation\changes.txt


hmailserver/source/Server/hMailServer.Minidump/hMailServer.Minidump.vcxproj
```diff
@@ -51,7 +51,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>$(hMailServerLibs)\boost_1_72_0;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(hMailServerLibs)\boost_1_73_0;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
martinknafve, line 54: This change has no relation to the "Installer bug fixes" right?

hmailserver/installation/section_files_64.iss
```diff
@@ -7,6 +7,11 @@ Source: "Microsoft.VC142.CRT\*"; DestDir: "{app}\Bin"; Flags: ignoreversion; Com
```
martinknafve, line 7: Is there some innosetup-magic which prevents this line from installing the Windows-*KB*-msu's into the bin-directory (and not just into the temp directory)?

martinknafve:
Thanks @Dravion 

I am thinking about whether we should be prompting the user before installing the KB*-executable. These are system-wide updates (right?) which may theoretically cause some issues in other software installed on the computer.

hmailserver/installation/section_files_64.iss
Dravion: > Is there some innosetup-magic which prevents this line from installing the Windows-_KB_-msu's into the bin-directory (and not just into the temp directory)?

No. Hmm, was thinking i changed it to Temp.
but of cause, we should change it to Temp!

Dravion:
> Thanks @Dravion
> 
> I am thinking about whether we should be prompting the user before installing the KB*-executable. These are system-wide updates (right?) which may theoretically cause some issues in other software installed on the computer.

More or less the same procedure as we have it before with distributing Redistrict stuff.

KB2999226 is just the Universal C Runtime stuff which i included into Innosetup
IMHO this is not a high impact Update, but  KB2919442 + KB2919355  (required for Win8.1 and Win2012_R only are high impact Updates.

KB2919355 alone can take up from 5 to 10 Minutes to install and it's around 707 MBytes
and needs a restart. Unfortunately, it is required to solve the "api-ms-win-crt-runtime-l1-1-0.dll"
Error on Win8.1 + Win2012_R2.

hmailserver/source/Server/hMailServer.Minidump/hMailServer.Minidump.vcxproj
Dravion: No! Sorry i was using a newer Boost version for my own test builds ;d

hmailserver/installation/section_files_64.iss
martinknafve: Do you intend to update the pull request with this or do you want me to fix it after merge?

martinknafve:
The VC++ redist files were placed into hMailServers bin directory, meaning that it did not do system-wide changes. These KB's on the other hand performs system-wide changes which may affect things outside of hMailServer, right? The installer should at least prompt the user to confirm that he want's to install these.

Dravion:
> The VC++ redist files were placed into hMailServers bin directory, meaning that it did not do system-wide changes. These KB's on the other hand performs system-wide changes which may affect things outside of hMailServer, right? The installer should at least prompt the user to confirm that he want's to install these.

The Problem is, if KB2999226 is NOT installed (on any OS below Win10 or Windows Server  2016)   hMailServer Installer will fail with Error message "api-ms-win-crt-runtime-l1-1-0.dll is missing" and
results in a corrupt hMailServer installation.

#Details regarding *.MSU files.
Windows6.1-KB2999226-x64.msu (covers Windows 7+SP1 and Win 2008_R2)
Windows8-RT-KB2999226-x64.msu (covers Windows 8 and Win Server 2012)
Windows8.1-KB2999226-x64.msu (covers Windows 8.1 and Win Server 2012_R2)

It's always the same UCrt Update, but any OS requires it's specific KB2999226 build!

#Extra problematic:
On Windows 8.1 and Windows Server 2012_R2, the required 
Windows8.1-KB2999226-x64.msu fails if  KB2919442  and KB2919355  is NOT installed. 
That's why i  check on Win 8.1 and Win 2012_R2 for KB2919442 and KB2919355 and
quit the hMailServer Installer if not found (see below)

The check is in hMailServerInnoExtension.iss / Line 495
`
if (Is_Win81_or_Win2012_R2()) and (IsKBInstalled('KB2919355') = false) then 	 
begin
 MsgBox('ERROR: hMailServer requires the following Windows Updates:  '+
 'Windows8.1-KB2919442-x64.msu AND Windows8.1-KB2919355-x64.msu.', mbError, MB_OK);
 ExitProcess(0);
end;
`

RvdHout:
Don't know if it is wise to include them KB2999226 packages at all (can those be silently updated?)
This way for a few OS's you include everything that is needed but for others you do not...kinda messy if you ask me. 

And after all we are talking about older systems here, what are the chances these systems are not up to date and already have all prerequisites installed? Why not simply display a message what is needed for the installer to run? Like you do with NetFX 4.5 and simply refuse to continue if one of the requirements is missing. You could put a link there that reference the KB needed, eg: https://support.microsoft.com/help/2999226/update-for-universal-c-runtime-in-windows . The prerequisites for KB2999226 are also listed there

Dravion:
Its only Win 8.1 or Windows Server 2012_R2 which has extra requirements to be able to install KB2999226 . If we are running Win10 or Windows Server 2016 or higher, the steps are not necessary.

RvdHout:
> Its only Win 8.1 or Windows Server 2012_R2 which has extra requirements to be able to install KB2999226 . If we are running Win10 or Windows Server 2016 or higher, the steps are not necessary.

I know, therefor why include KB2999226? What are the chances hmailserver is installed on a new/clean (read: not up to date) older OS like Server 2012 or Server 2012 R2? Especially when a few OS's, like Server 2012 R2 need additional KB's? 

I really don't see the point to included them KB2999226 packages in the installer, making it **unnecessary bigger** for perhaps 1% of the users that try to install it on a clean, **not updated** older OS that are almost EOL

Dravion:
> > Its only Win 8.1 or Windows Server 2012_R2 which has extra requirements to be able to install KB2999226 . If we are running Win10 or Windows Server 2016 or higher, the steps are not necessary.
> 
> I know, therefor why include KB2999226? What are the chances hmailserver is installed on a new/clean (read: not up to date) older OS like Server 2012 or Server 2012 R2? Especially when a few OS's, like Server 2012 R2 need additional KB's?
> 
> I really don't see the point to included them KB2999226 packages in the installer, making it **unnecessary bigger** for perhaps 1% of the users that try to install it on a clean, **not updated** older OS that are almost EOL

According to Microsoft Support, Windows Server 2012 and Windows Server 2012_R2
has extended support til 10/10/2023
Reference:
https://support.microsoft.com/en-us/lifecycle/search?alpha=Windows%20Server%202012

RvdHout:
> > > Its only Win 8.1 or Windows Server 2012_R2 which has extra requirements to be able to install KB2999226 . If we are running Win10 or Windows Server 2016 or higher, the steps are not necessary.
> > 
> > 
> > I know, therefor why include KB2999226? What are the chances hmailserver is installed on a new/clean (read: not up to date) older OS like Server 2012 or Server 2012 R2? Especially when a few OS's, like Server 2012 R2 need additional KB's?
> > I really don't see the point to included them KB2999226 packages in the installer, making it **unnecessary bigger** for perhaps 1% of the users that try to install it on a clean, **not updated** older OS that are almost EOL
> 
> According to Microsoft Support, Windows Server 2012 and Windows Server 2012_R2
> has extended support til 10/10/2023
> Reference:
> https://support.microsoft.com/en-us/lifecycle/search?alpha=Windows%20Server%202012

So? Does that significantly increase the possibility hmailserver is installed on a clean, not up to date version of that OS?

Dravion:
> > > > Its only Win 8.1 or Windows Server 2012_R2 which has extra requirements to be able to install KB2999226 . If we are running Win10 or Windows Server 2016 or higher, the steps are not necessary.
> > > 
> > > 
> > > I know, therefor why include KB2999226? What are the chances hmailserver is installed on a new/clean (read: not up to date) older OS like Server 2012 or Server 2012 R2? Especially when a few OS's, like Server 2012 R2 need additional KB's?
> > > I really don't see the point to included them KB2999226 packages in the installer, making it **unnecessary bigger** for perhaps 1% of the users that try to install it on a clean, **not updated** older OS that are almost EOL
> > 
> > 
> > According to Microsoft Support, Windows Server 2012 and Windows Server 2012_R2
> > has extended support til 10/10/2023
> > Reference:
> > https://support.microsoft.com/en-us/lifecycle/search?alpha=Windows%20Server%202012
> 
> So? Does that significantly increase the possibility hmailserver is installed on a clean, not up to date version of that OS?

We can show only warnings regarding required KB Updates, quit and leave it to the User.
But i think we will see a few more support cases in the forum because of this, 

I can't count the number of libmysql.dll topics in Forum  anymore...

RvdHout:
libmysql.dll is a whole different matter, as for that it doesn't matter if it a old OS or a new like Windows 10 and Server 2019... and therefor libmysql.dll should be included

martinknafve:
I'm thinking about another solution to this issue. 

hMailServer installation already bundle C++ runtime files and put them into Bin directory. These DLL's depend on Universal C Runtime (for example api-ms-win-core-memory-l1-1-0.dll) which comes preinstalled on Windows 10 but not on some earlier versions. On earlier versions, there are two ways to solve it: Either install the KB items as in this commit, or ship the files as part of the installation. Both options are supported by Microsoft. Since the hMailServer installer already ship the C++ runtime DLL's (rather than relying on a specific KB to be installed), it seems consistent it would also ship the Universal C Runtime DLL's (since the C++ runtime DLL's has a direct dependency on them).

I think just shipping these DLL's would be a less surprising / scary behavior than automatically installing a windows update which is several hundred megabytes which may cause issues for other software running on the same machine. 

Shipping the DLLs increase installation size with only about 400KB.

Do you see an issue with that approach?

RvdHout:
that should need some testing, especially on OS's that would normally require additional KB's ... does including the Universal C Runtime DLL's behave as wanted if a prerequisites is missing

Dravion:
> I'm thinking about another solution to this issue.

> I think just shipping these DLL's would be a less surprising / scary behavior than automatically installing a windows update which is several hundred megabytes which may cause issues for other software running on the same machine.

Not so fast @Martin!

The commit includes **ONLY** 
Windows6.1-KB2999226-x64.msu
Windows8.1-KB2999226-x64.msu
Windows8-RT-KB2999226-x64.msu

Which will increase the hMailServer Installer 3,2 MBytes in total.
Before i pushed the commit as Pull request, i made a Demo Installer 
which includes everything. 

You can download it and test it in one of your VM's right now to see it in action
https://github.com/Dravion/hmailserver/releases

RvdHout:
Geez @Dravion , still a problem reading what he is actually saying? @martinknafve is talking about DDL's

@martinknafve, afraid every OS needs is specific set of DDL's, version wise?

martinknafve:
Okay @Dravion, but why would we want 2 separate mechanisms for distributing C++ Runtime and Universal C Runtime when they are the exact same type of dependency? (Local deployment for one, and KB-install for another?).

RvdHout:
> Only on Win 8.1 and Windows Server 2012_R2 the User gets an ERROR-Message that a
> ADDITIONAL required update (in order to Install the Universal Runtime KB2999226)
> 
> KB2919442 + KB2919355 needs to be manually downloaded and installed first.
> The KB2919355 is about 707 MBytes big, but AGAIN, not INCLUDED in the Installer,
> or automatically downloaded and silently installed!

And that is exactly why I said that is messy...full support for some, no full support for others.... you cannot seriously consider that a smooth installation experience

RvdHout:
> @martinknafve, afraid every OS needs is specific set of DLL's, version wise?

i compared the Server 2012 msu with the Server 2012 R2 msu, DLL version info seems te be matching, eg: 10.0.10240.16930 for api-ms-win-crt-runtime-l1-1-0.dll (x64) has same hashes in both

CRC32: 191003D2
MD5: BBAE7B5436D6D1B0FC967FF67E35415F
SHA-1: F67BC165CEFB119AD767B6BEC27A1102C0FD2BAC

Dravion:
> I'm thinking about another solution to this issue.
> 
> hMailServer installation already bundle C++ runtime files and put them into Bin directory. These DLL's depend on Universal C Runtime (for example api-ms-win-core-memory-l1-1-0.dll) which comes preinstalled on Windows 10 but not on some earlier versions. On earlier versions, there are two ways to solve it: Either install the KB items as in this commit, or ship the files as part of the installation. Both options are supported by Microsoft. Since the hMailServer installer already ship the C++ runtime DLL's (rather than relying on a specific KB to be installed), it seems consistent it would also ship the Universal C Runtime DLL's (since the C++ runtime DLL's has a direct dependency on them).
> 
> I think just shipping these DLL's would be a less surprising / scary behavior than automatically installing a windows update which is several hundred megabytes which may cause issues for other software running on the same machine.
> 
> Shipping the DLLs increase installation size with only about 400KB.
> 
> Do you see an issue with that approach?

**Sadly YES!**

This was my first approach to fix this thing, but it didn't work out, to my regret.

The Problem is, any KB2999226 Package targets a specific Window Version 
See the list of Downloads below:
https://support.microsoft.com/en-us/help/2999226/update-for-universal-c-runtime-in-windows

Next problem is the nature of this Universal Runtime Update thing

You can unpack a Windows Update *.msu file by using the Windows Expand command:

for example:
mkdir Win6.3 | expand –f:* Windows8.1-KB2999226-x64.msu Win6.3

To unpack the desired DLL's, use expand again:
mkdir DLLs|expand Windows8.1-KB2999226-x64.cab -F:* DLLs

This will extract 45 files, most of the DLL's! 

And now the real pain just starts to begin:

There are **3.x versions of api-ms-win-crt-runtime-l1-1-0.dll**, targeting 
specific folders within C:\Windows

For example:
C:\Windows\System32\api-ms-win-crt-runtime-l1-1-0.dll
C:\Windows\SysWOW64\api-ms-win-crt-runtime-l1-1-0.dll
C:\Windows\WinSxS\api-ms-win-crt-runtime-l1-1-0.dll

...and all the 15 also required DLL's like 
api-ms-win-crt-conio-l1-1-0.dll
api-ms-win-crt-filesystem-l1-1-0.dll
api-ms-win-crt-heap-l1-1-0.dll
...ect

This means, if you try to do it manually, you have to handle
*45 DLL's alone For Windows 7+SP1/Win2008_R2
*45 DLL's alone For Windows 8.0+SP1/Win2012
*45 DLL's alone For Windows 8.1+SP1/Win2012_R2

and the kicker is, you can't place them just like msvcp140.dll inside hMailServer\bin
directory. If you try this results in two type of Errors:

1). Entry point not found
or
2) The Application cannot initialilze

At this point i gave up this bug fix and simply added the 3.x, small *.msu files
into the Innosetup installer and it works.

See the Demo Installer:
https://github.com/Dravion/hmailserver/releases/download/5.7.0-B2519/hMailServer-5.7.0-B2519-x64.exe

RvdHout:
that doesn't sound right, how can the hashes between dll's in the different OS msu packages be the same then?

Dravion:
> that doesn't sound right, how can the hashes between dll's in the different OS msu packages be the same then? We do not need the 32-bit dll's, 5.7 is 64-bit only

I don't know
I spent a whole day to setup up different VM's from Win7 up to Win10/Win2019
(clean install without any updates or Service packs)  and this was the result. 

Martin is right. 
This would be the easiest and most straight forward approach i can think of,
 but sadly it doesn't work.

RvdHout:
> Martin is right.
> This would be the easiest and most straight forward approach i can think of,
> but sadly it doesn't work.

says you.... reason enough to give it a another try

martinknafve:
> This would be the easiest and most straight forward approach i can think of, but sadly it doesn't work.

What issue did it cause?

Dravion:
> > This would be the easiest and most straight forward approach i can think of, but sadly it doesn't work.
> 
> What issue did it cause?

**THIS:**

This was my first approach to fix this thing, but it didn't work out, to my regret.

The Problem is, any KB2999226 Package targets a specific Window Version
See the list of Downloads below:
https://support.microsoft.com/en-us/help/2999226/update-for-universal-c-runtime-in-windows

Next problem is the nature of this Universal Runtime Update thing

You can unpack a Windows Update *.msu file by using the Windows Expand command:

for example:
mkdir Win6.3 | expand –f:* Windows8.1-KB2999226-x64.msu Win6.3

To unpack the desired DLL's, use expand again:
mkdir DLLs|expand Windows8.1-KB2999226-x64.cab -F:* DLLs

This will extract 45 files, most of the DLL's!

And now the real pain just starts to begin:

There are 3.x versions of api-ms-win-crt-runtime-l1-1-0.dll, targeting
specific folders within C:\Windows

For example:
C:\Windows\System32\api-ms-win-crt-runtime-l1-1-0.dll
C:\Windows\SysWOW64\api-ms-win-crt-runtime-l1-1-0.dll
C:\Windows\WinSxS\api-ms-win-crt-runtime-l1-1-0.dll

...and all the 15 also required DLL's like
api-ms-win-crt-conio-l1-1-0.dll
api-ms-win-crt-filesystem-l1-1-0.dll
api-ms-win-crt-heap-l1-1-0.dll
...ect

This means, if you try to do it manually, you have to handle
*45 DLL's alone For Windows 7+SP1/Win2008_R2
*45 DLL's alone For Windows 8.0+SP1/Win2012
*45 DLL's alone For Windows 8.1+SP1/Win2012_R2

and the kicker is, you can't place them just like msvcp140.dll inside hMailServer\bin
directory. If you try this results in two type of Errors:

1). Entry point not found
or
2) The Application cannot initialilze

At this point i gave up and simply added the 3.x, small *.msu files
into the Innosetup installer and it works as you can see in my Demo installer
(which reflects the Commit)

Demo Installer download:
https://github.com/Dravion/hmailserver/releases/download/5.7.0-B2519/hMailServer-5.7.0-B2519-x64.exe

RvdHout:
@dravion, are you sure you didn't screw up? you mention SysWoW, WinSxS, and System32 dll's...but just one, only the 64-bit files matter, the rest as well as the info you keep coping and pasting is not even relevant

martinknafve:
I also got the same when doing the same thing. I will troubleshoot that.

RvdHout:
> I also got the same when doing the same thing. I will troubleshoot that.

also included ucrtbase.dll?

RvdHout:
https://docs.microsoft.com/nl-nl/cpp/windows/universal-crt-deployment?view=vs-2019

martinknafve:
> https://docs.microsoft.com/cpp/windows/universal-crt-deployment?view=vs-2019

Yes I read that one.

> "The set of DLLs required on each operating system varies."

Not sure if this means that specific DLL's versions vary from OS to OS, or if it means that on some OS some of the DLL's are required and on other OS'es other DLL's are required.

>> I also got the same when doing the same thing. I will troubleshoot that.
> also included ucrtbase.dll?

Yes. It worked fine in Windows Server 2008 R2 but not in Windows Server 2008 SP2.

RvdHout:
> > https://docs.microsoft.com/cpp/windows/universal-crt-deployment?view=vs-2019
> 
> Yes I read that one.
> 
> > "The set of DLLs required on each operating system varies."
> 
> Not sure if this means that specific DLL's versions vary from OS to OS, or if it means that on some OS some of the DLL's are required and on other OS'es other DLL's are required.

Did you use "Windows Kits\10\Redist\ucrt\DLLs"? 
I 'think' they mean by the set of DLLs required there might be additional files, eg:

Windows8-RT-KB2999226 = 17 + ucrtbase.dll  = 18 dll's
Windows8.1-KB2999226 = 15 + ucrtbase.dll = 16 dll's

Dravion:
> > https://docs.microsoft.com/cpp/windows/universal-crt-deployment?view=vs-2019
> 
> Yes I read that one.
> 
> > "The set of DLLs required on each operating system varies."
> 
> Not sure if this means that specific DLL's versions vary from OS to OS, or if it means that on some OS some of the DLL's are required and on other OS'es other DLL's are required.
> 
> > > I also got the same when doing the same thing. I will troubleshoot that.
> > > also included ucrtbase.dll?
> 
> Yes. It worked fine in Windows Server 2008 R2 but not in Windows Server 2008 SP2.

This is because the DLL's are build for specific OS Major+Minor version numbers
Like this:
6.0 (Vista and Windows Server 2008)
6,1 (Win7 and Windows Server 2008 R2)
6.2 (Win8 and Windows Server 2012)
6.3 (Win8.1 and Windows Server 2012 R2)

RvdHout:
> > > https://docs.microsoft.com/cpp/windows/universal-crt-deployment?view=vs-2019
> > 
> > 
> > Yes I read that one.
> > > "The set of DLLs required on each operating system varies."
> > 
> > 
> > Not sure if this means that specific DLL's versions vary from OS to OS, or if it means that on some OS some of the DLL's are required and on other OS'es other DLL's are required.
> > > > I also got the same when doing the same thing. I will troubleshoot that.
> > > > also included ucrtbase.dll?
> > 
> > 
> > Yes. It worked fine in Windows Server 2008 R2 but not in Windows Server 2008 SP2.
> 
> This is because the DLL's are build for specific OS Major+Minor version numbers
> Like this:
> 6.0 (Vista and Windows Server 2008)
> 6,1 (Win7 and Windows Server 2008 R2)
> 6.2 (Win8 and Windows Server 2012)
> 6.3 (Win8.1 and Windows Server 2012 R2)

You are so wrong.... read the document linked to above, and especially the part for "Local deployment"

martinknafve:
> Did you use "Windows Kits\10\Redist\ucrt\DLLs"? I suspect this will include them all, not?

Yes, I included all of them with the installer. 

> I 'think' they mean by the set of DLLs required there might be additional files, eg:

That's my interpretation as well. But still gives error on Windows Server 2008.

"Windows Kits\10" is referring to the Windows SDK folder structure. The latest Windows SDK (which I'm also using) only supports Windows Server 2012 and later. Since UCRT is relatively new, I doubt there's a Windows SDK version which includes a Windows Server 2008-compatible Redist-version of UCRT.

I do wonder if the Windows Server 2008 x64 UCRT-version also works on later versions of Windows automatically though.

RvdHout:
the initial version of the Universal CRT (version 10.0.10240) 
https://developer.microsoft.com/nl-nl/windows/downloads/sdk-archive/

But dunno if that one will support 2008/Vista....

Dravion:
> > Did you use "Windows Kits\10\Redist\ucrt\DLLs"? I suspect this will include them all, not?
> 
> Yes, I included all of them with the installer.
> 
> > I 'think' they mean by the set of DLLs required there might be additional files, eg:
> 
> That's my interpretation as well. But still gives error on Windows Server 2008.

You will have the same  Error for Windows 8 / 2012 AND Win 8.1 / 2012_R2 as well.

RvdHout:
> > > Did you use "Windows Kits\10\Redist\ucrt\DLLs"? I suspect this will include them all, not?
> > 
> > 
> > Yes, I included all of them with the installer.
> > > I 'think' they mean by the set of DLLs required there might be additional files, eg:
> > 
> > 
> > That's my interpretation as well. But still gives error on Windows Server 2008.
> 
> You will have the same Error for Windows 8 / 2012 AND Win 8.1 / 2012_R2 as well.

Universal CRT is independent of OS, hence...that's why it is called **Universal**
But please, keep being a twat and stick with your truth 

> The Universal CRT is a Microsoft Windows operating system component included as part of the operating system in Windows 10. It's available for older operating systems, Windows Vista through Windows 8.1, by using Windows Update. Local deployment of the Universal CRT is supported, with some restrictions.

> On versions of Windows before Windows 8, the Universal CRT can't be packaged locally with a plugin, if it's located in a directory other than the directory of the main app executable. The APISet forwarder DLLs are unable to resolve the ucrtbase.dll successfully in this case. Some recommended alternative solutions include:
> 
> Statically link the Universal CRT,
> Centrally deploy the Universal CRT, or
> Place the Universal CRT files in the same directory as the app.

Dravion:
The DLL's from C:\Program Files (x86)\Windows Kits\10\Redist\10.0.18362.0\ucrt\DLLs\x64
are working for Win 7 SP1 + Win 2008 R2 up Windows 8 / Windows 8.1 and Server 2012/R2.

In this case, you solution works and is better as my central deployment solution, installing
KB Updates.

PR #337, opened 2020-05-14
Set JScript engine version

thatoddmailbox:
This pull request adds some code to request version 5.8 of the JScript engine from Windows for scripts. The main reason I did this is to get access to built-in JSON object parsing, but there are several other useful features (like being able to index through strings!) that you can see here: https://docs.microsoft.com/en-us/archive/blogs/jscript/versioning-language-features-in-jscript (scroll down to "List of features versioned")

I don't think any of the new features in version 5.8 are incompatible with existing scripts, so I have it enabling the new version for everything.

I also fixed a minor bug with the OnExternalAccountDownload event and how it called JScript code when the remote UID had a single quote in it.

### DIFF
--- a/hmailserver/source/Server/Common/Scripting/Events.cpp
+++ b/hmailserver/source/Server/Common/Scripting/Events.cpp
@@ -186,7 +186,7 @@ namespace HM
       }
       else if (sScriptLanguage == _T("JScript"))
       {
-         sRemoteUIDCopy.Replace(_T("'"), _T("\'"));
+         sRemoteUIDCopy.Replace(_T("'"), _T("\\'"));
 
          if (pMessage)
             sEventCaller.Format(_T("OnExternalAccountDownload(HMAILSERVER_FETCHACCOUNT, HMAILSERVER_MESSAGE, '%s')"), sRemoteUIDCopy.c_str());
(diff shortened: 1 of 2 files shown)


martinknafve:
@thatoddmailbox I had to revert the code which set script version to 5.8. Turns out this didn't work on Windows Server 2008 (I think the 5.8 engine version was introduced later). I'm thinking if we should just have some code which checks if user is running Windows Server 2008  and if so skip this part. What do you think?

thatoddmailbox:
Oh, whoops! Looking into it some more I think that it's actually connected to the version of Internet Explorer you have installed. If you have IE 8 or newer, then you have JScript 5.8, otherwise it doesn't work. The newer versions of Windows and Windows Server come with newer version of IE, which is why it works.

I think it is reasonable to fall back to the the older version if it's required, although it does feel a little weird to do that to the administrator, if features are disabled or enabled depending on what they have as their IE or Windows version. I'm not sure if there is a better solution though.

RvdHout:
Isn't the script engine 5.8 installed with IE8? But I don't understand where this check came from... you can use what is available to you, either being 5.7 or 5.8 script engine, not?

PR #338, opened 2020-05-14
Add OnClientValidatePassword event

thatoddmailbox:
This pull request adds a new `OnClientValidatePassword` event for scripts. The idea is to allow scripts to authenticate users. For example, if you want to validate user passwords against a database or external login provider, you could write a script to handle that event.

The event is called with two parameters (`OnClientValidatePassword(oAccount, sPassword)`): the account that is trying to sign in and the password that's being used. The script can then set Result.Value to:
* 0 - The user was authenticated, let them in.
* 1 - The user was not authenticated, return an "invalid username or password" error.
* 2 - _(default)_ Bypass the script and continue with the normal hMailServer password validation.

Hopefully this makes some sense, let me know if you have any feedback!

(also, is there any way to add to the documentation [on this page](https://www.hmailserver.com/documentation/latest/?page=reference_scripts)? I wanted to add a page about this event as part of the pull request but couldn't find where that comes from)

### DIFF
--- a/hmailserver/source/Server/Common/Scripting/ScriptObjectContainer.cpp
+++ b/hmailserver/source/Server/Common/Scripting/ScriptObjectContainer.cpp
@@ -10,6 +10,7 @@
 #include "../../COM/InterfaceClient.h"
 #include "../../COM/InterfaceEventLog.h"
 #include "../../COM/InterfaceFetchAccount.h"
+#include "../../COM/InterfaceAccount.h"
 
 //#include "IScriptObject.h"
 #include "ScriptObjectContainer.h"
@@ -100,6 +101,15 @@ namespace HM
          {
             CComObject<InterfaceFetchAccount> *pInterface = new CComObject<InterfaceFetchAccount>();
             std::shared_ptr<FetchAccount> pObject = std::static_pointer_cast<FetchAccount>(pObj->pObject);
+            pInterface->AttachItem(pObject);
+            pInterface->QueryInterface(ppunkItem);
+
+            return true;
+         }
+      case ScriptObject::OTAccount:
+         {
+            CComObject<InterfaceAccount> *pInterface = new CComObject<InterfaceAccount>();
+            std::shared_ptr<Account> pObject = std::static_pointer_cast<Account>(pObj->pObject);
             pInterface->AttachItem(pObject);
             pInterface->QueryInterface(ppunkItem);
 
--- a/hmailserver/source/Server/Common/Scripting/ScriptObjectContainer.h
+++ b/hmailserver/source/Server/Common/Scripting/ScriptObjectContainer.h
@@ -17,7 +17,8 @@ namespace HM
          OTMessage = 1001,
          OTClient = 1002,
          OTEventLog = 1003,
-         OTFetchAccount = 1004
+         OTFetchAccount = 1004,
+         OTAccount = 1005
       };
 
       String sName;
--- a/hmailserver/source/Server/Common/Scripting/ScriptServer.cpp
+++ b/hmailserver/source/Server/Common/Scripting/ScriptServer.cpp
@@ -101,6 +101,7 @@ namespace HM
          has_on_smtpdata_ = DoesFunctionExist_("OnSMTPData");
          has_on_helo_ = DoesFunctionExist_("OnHELO");
          has_on_client_logon_ = DoesFunctionExist_("OnClientLogon");
+         has_on_client_validate_password_ = DoesFunctionExist_("OnClientValidatePassword");
       }
       catch (...)
       {
@@ -262,6 +263,11 @@ namespace HM
          if (!has_on_client_logon_)
             return;
          event_name = _T("OnClientLogon");
+         break;
+      case EventOnClientValidatePassword:
+         if (!has_on_client_validate_password_)
+            return;
+         event_name = _T("OnClientValidatePassword");
          break;
       case EventCustom:
          break;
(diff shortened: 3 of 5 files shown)


FabulousGee:
@thatoddmailbox love this! I will see if I am able to set up a build environment the next days as I would love to test this out. This would be a neat feature since we could finally add an easy way to allow different auth-methods (for me this will be LDAP).

@martinknafve maybe you can test this quicker and just integrate the pull-request?

FabulousGee:
I can confirm this working. I built my own version of hmailserver depending on hMailServer 5.6.7 - Build 2425, so I can say it's running smooth on the latest stable and may be added there. Later versions have to be still confirmed working without flaws.

RvdHout:
I think you forgot to set the default value,

ScriptServer.cpp, line 35 add:
has_on_client_validate_password_(false),

thatoddmailbox:
@RvdHout Good catch, sorry about that! It should be fixed now.

martinknafve:
I did two more changes to this after merge:
* I moved the logic to call the event from PasswordValidator to Events class. Similar logic is already in Events class and I prefer to have the scripting code there: https://github.com/hmailserver/hmailserver/commit/c3738c424c923518872cc47b3f4e1f65b249d2ae
* I added an integration test which tests the new event: https://github.com/hmailserver/hmailserver/commit/ea5066b5984281a16db28f0b5d83291274d6d325

RvdHout:
@martinknafve 

I think this commit should be revoked immediately, this is really, really dangerous!

If someone (hacker) or malware has obtained access to the server they can login to every account by simply changing a single line of code

```
Sub OnClientValidatePassword(oAccount, sPassword)
	Result.Value = 0
End Sub
```

FabulousGee:
@RvdHout but if anyone has obtained access to the server, wouldn't he have access to everything anyway?
Pretty sure that in this situation the server is lost anyway, isn't it?

RvdHout:
@FabulousGee 

No, i don't fully agree... they still additional work done, eg: need to decrypt the password and stuff
With this line of code they won't need anything like that

FabulousGee:
@RvdHout
I am not sure if we did something wrong on our end but would you mind having a look at the following folder(s)?
C:\Program Files (x86)\hMailServer\Data\YOUR_DOMAIN\YOUR_USERNAME\...

At our end, there is a whole lot of folders in there which hold all emails in plain text. So if someone would have access to our server he could read any email this way?! I know some other mailservers on linux doing this as well.

So, again, except for the password itself, I don't see any security obtained without the hook.
We could discuss if the password is important enough to think of removing the hook but since I do not know the other parameters especially regarding the strength of the encryption used I will take no position on that one so far

RvdHout:
I said i didn't fully agree, sure if they physically obtained access you are screwed anyway.
But with this one liner in OnClientValidatePassword they not only can access your mail, they also can send out mail on behalf of clients without having to decrypt/sniff for password. (which is more likely to be noticed)

SorenRR:
@FabulousGee , @RvdHout
If someone breaches security on server they cannot get to the passwords and decode them in less than 150 years, no need to make it any easier for them.

The functionality of OnClientValidatePassword in it's current form is - and I agree with Ruud here - potentially very dangerous. Obtaining access using no password violate everything I have learned about IT security over the past 35 years!

Also the fact that a password can be collected in clear text and associated with a client account is also questionable in a public set-up. Ideally the password should be hashed/encrypted and simply compared with external authentication holding a shared key for equally hashed/encrypted passwords. Only if the hashed/encrypted passwords compare should a positive response be returned.

FabulousGee:
@SorenRR decoding in about 150 years? What algorithm is currently being used there? Do you have it in mind currently?

While I still question the benefit of having/gaining a breached/clear-text password nowadays in general (that's what I wanted to say with the quoted article)
on the other hand I have to admit that I also like the idea of comparing hashes more than the current way. I don't think that it will increase security severely but it's clearly best-practise and we really should not make it easier to get to the passwords than necessary (no matter if it's a small, middle or big security flaw in the end).

SorenRR:
@FabulousGee I take my passwords with some salted SHA256 ;-)

Question is how long time do you expect intruders will spend trying to hack a password? Salted MD5 = 2-3 hours on a normal PC. Salted SHA1 could take a couple of weeks of supercomputer time. If we extend hashing with bCrypt or PBKDF2 then passwords will be virtually unbreakable.

I'm not into the legalities of EU GDPR but I suspect easy access to passwords is not looked at kindly.

OutbackMatt:
> @RvdHout but if anyone has obtained access to the server, wouldn't he have access to everything anyway?
> Pretty sure that in this situation the server is lost, isn't it?

Absolutely,
However finding this this particular little needle in a hay stack would be near impossible

We used to have passwords in plain text in the logs, and we removed that years ago.
This new event makes this trivial



> @martinknafve
> 
> I think this commit should be revoked immediately as this is really, really dangerous!
> 

I agree

RvdHout:
Not to forget physical access is not our only concern. Remember the solorwinds hack? Petya? Those malware/ransomware came on to systems using (modified) legitimate software tools... 

It's not a hard task for malware builders to build in a check that looks if hMailServer is installed, and if so find the EventHandlers.vbs and adjust a few lines.

I remember something like this happened to me years back, OK i have to admit the source op the program was questionable then, but all of the sudden some malware within was trying to send all my FileZilla credentials (stored in sitemanager) over the internet. Luckily my firewall stopped the outgoing traffic and i was able to remove the malware without leaking this valuable info.
C:\Users\[USERNAME]\AppData\Roaming\FileZilla (sitemanager.xml, base64 encoded passwords)

Issue #63, opened 2014-11-24
Is there a STARTTLS Vulnerability?


UweDD2:
I've installed newest release of hmailserver. A scan by the Greenbone Security Assistant revealed/reported a STARTTLS Vulnerability.
"Multiple Vendors STARTTLS Implementation Plaintext Arbitrary Command Injection Vulnerability
Summary:
An attacker can exploit this issue to execute arbitrary commands in the context of the user running the application. Successful exploits can allow attackers to obtain email usernames and passwords."

Link:
http://www.kb.cert.org/vuls/id/MAPG-8D9M75

I haven't found anything about it for hmailserver. But there are many hotfixes for other programs (postfix, ...). Is it a real bug or just a false positive.

There is also a topic at the user forum: 
https://www.hmailserver.com/forum/viewtopic.php?f=10&t=27395

PR #342, opened 2020-07-20
Fix 'Unencrypted Cleartext Login' vulnerability detection

uhliksk:
**hMailServer: 5.6.8 B2501 (SMTP+POP3), 5.7 (SMTP+POP3+IMAP)**

**Problem:**

Problem was already discussed in #63 marked as solved but still present. If you enable "STARTTLS (Required)" for SMTP/POP3/IMAP connections it'll be detected as 'Unencrypted Cleartext Login' vulnerability because hmailserver will return AUTH/USER/LOGIN as possible capability even if it is not possible to use login before TLS connection is established.

**Solution:**

Don't send AUTH/USER/LOGIN in capabilities list before TLS connection is established if TLS is required.

**Note:**
IMAP is affected if "SASL Plain" is enabled. SMTP and POP3 are affected in default configuration.

### DIFF
--- a/hmailserver/source/Server/IMAP/IMAPCommandCapability.cpp
+++ b/hmailserver/source/Server/IMAP/IMAPCommandCapability.cpp
@@ -36,7 +36,7 @@ namespace HM
           pConnection->GetConnectionSecurity() == CSSTARTTLSRequired)
          sResponse += " STARTTLS";
 
-      if (pConfig->GetUseIMAPSASLPlain())
+      if (pConfig->GetUseIMAPSASLPlain() && (pConnection->IsSSLConnection() || pConnection->GetConnectionSecurity() != CSSTARTTLSRequired))
 	      sResponse += " AUTH=PLAIN";
 
       if (pConfig->GetUseIMAPSASLInitialResponse())
--- a/hmailserver/source/Server/POP3/POP3Connection.cpp
+++ b/hmailserver/source/Server/POP3/POP3Connection.cpp
@@ -375,7 +375,10 @@ namespace HM
    void
    POP3Connection::ProtocolCAPA_()
    {
-      String capabilities = "USER\r\nUIDL\r\nTOP\r\n";
+      String capabilities = "UIDL\r\nTOP\r\n";
+
+      if (IsSSLConnection() || GetConnectionSecurity() != CSSTARTTLSRequired)
+         capabilities+="USER\r\n";
 
       if (GetConnectionSecurity() == CSSTARTTLSOptional ||
           GetConnectionSecurity() == CSSTARTTLSRequired)
--- a/hmailserver/source/Server/SMTP/SMTPConnection.cpp
+++ b/hmailserver/source/Server/SMTP/SMTPConnection.cpp
@@ -1467,7 +1467,7 @@ namespace HM
          }
       }
 
-      if (GetAuthIsEnabled_())
+      if (GetAuthIsEnabled_() && (IsSSLConnection() || GetConnectionSecurity() != CSSTARTTLSRequired))
       {
          String sAuth = "\r\n250-AUTH LOGIN";
 


RvdHout:
But is this a 'real' vulnerability, as you won't be able to login encrypted when STARTTLS is required?
Your fix strips them, AUTH/USER/LOGIN out completely, right? Doesn't such a connection need to know the capabilities?

@martinknafve, thoughts?

uhliksk:
> But is this a 'real' vulnerability, as you won't be able to login unencrypted when STARTTLS is required?
> Your fix strips them, AUTH/USER/LOGIN out completely, right? Doesn't such a connection need to know the capabilities?
> 
> @martinknafve, thoughts?

It's not vulnerable because in reality it will not allow to login without STARTTLS. It's just false detected by OpenVAS as vulnerability because of false claim by hmailserver AUTH/USER/LOGIN are possible before STARTTLS even when they're not. Server should never report capabilities which is in reality not capable in current state.

Note: Hmailserver is reporting "503 must have sender first." to RCPT TO if STARTTLS is required. It should report "530 Must issue STARTTLS first." instead. It's just opening possible attack vector going thru piece of code it should never go thru if STARTTLS is required. Not good, not terrible, but better to fix also.

uhliksk:
Just for reference - RFC3207 - https://tools.ietf.org/html/rfc3207

- The server MUST discard any knowledge obtained from the client, such as the argument to the EHLO command, which was not obtained from the TLS negotiation itself.  The client MUST discard any knowledge obtained from the server, such as the list of SMTP service extensions, which was not obtained from the TLS negotiation itself.  The client SHOULD send an EHLO command as the first command after a successful TLS negotiation.
- A SMTP server that is not publicly referenced may choose to require that the client perform a TLS negotiation before accepting any commands.  In this case, the server SHOULD return the reply code:
    530 Must issue a STARTTLS command first
to every command other than NOOP, EHLO, STARTTLS, or QUIT.  If the client and server are using the ENHANCEDSTATUSCODES ESMTP extension [RFC2034], the status code to be returned SHOULD be 5.7.0.

_

uhliksk:
@RvdHout As you can see, client should not rely on capabilities list obtained before STARTTLS and should send EHLO again after TLS is started to get actual capabilities list.

RvdHout:
> Note: Hmailserver is reporting "503 must have sender first." to RCPT TO if STARTTLS is required. It should report "530 Must issue STARTTLS first." instead. It's just opening possible attack vector going thru piece of code it should never go thru if STARTTLS is required. Not good, not terrible, but better to fix also.

Uh, that isn't to hard.  SMTPConnection::ProtocolRCPT_(const String &Request)

```
      if (!IsSSLConnection() && GetConnectionSecurity() == CSSTARTTLSRequired)
      {
         EnqueueWrite_("530 must issue a STARTTLS command first.");
         return;
      }
```


```
"SMTPD"	8456	21	"2020-12-21 22:43:10.231"	"xxx.xxx.xxx.xxx"	"SENT: 220 mail.vdhout.nl ESMTP"
"SMTPD"	10504	21	"2020-12-21 22:43:24.891"	"xxx.xxx.xxx.xxx"	"RECEIVED: EHLO gmail.com"
"DEBUG"	10504	"2020-12-21 22:43:24.891"	"Executing event OnHELO"
"DEBUG"	10504	"2020-12-21 22:43:25.016"	"Event completed"
"SMTPD"	10504	21	"2020-12-21 22:43:25.016"	"xxx.xxx.xxx.xxx"	"SENT: 250-mail.domain.com[nl]250-SIZE 51200000[nl]250-STARTTLS[nl]250 HELP"
"SMTPD"	8456	21	"2020-12-21 22:43:35.921"	"xxx.xxx.xxx.xxx"	"RECEIVED: rcpt to: yy@gmail.com"
"SMTPD"	8456	21	"2020-12-21 22:43:35.937"	"xxx.xxx.xxx.xxx"	"SENT: 530 must issue a STARTTLS command first."

RvdHout:
That was a quick and dirty test... 
There is already a function,  CheckStartTlsRequired_()

What you are saying this should be called before MAIL(*), RCPT, DATA, AUTH(*)
CheckStartTlsRequired_() was already being called for command types marked with *

What about ETRN?

RvdHout:
@uhliksk, it seems you created a pull request for 5.7.x branche, not the 5.6.8 branche you mentioned in the first post of this topic, eg: [IMAPCommandCapability.cpp](https://github.com/hmailserver/hmailserver/blob/5.6.8/hmailserver/source/Server/IMAP/IMAPCommandCapability.cpp#L35) is quite different and missing the SASL stuff

uhliksk:
> @uhliksk, it seems you created a pull request for 5.7.x branche, not the 5.6.8 branche you mentioned in the first post of this topic, eg: [IMAPCommandCapability.cpp](https://github.com/hmailserver/hmailserver/blob/5.6.8/hmailserver/source/Server/IMAP/IMAPCommandCapability.cpp#L35) is quite different and missing the SASL stuff

Yes, sorry for confusion, first I found a bug in 5.6.8 for SMTP and POP3 and then I discovered same problem for IMAP when checking code in master branch so I decided to make PR into master branch to fix all three problems before next version is released. I forgot to change version number in post which I had already prepared. My mistake.

uhliksk:
> There is already a function, [CheckStartTlsRequired_()](https://github.com/hmailserver/hmailserver/blob/e3f91404330fc5cadcc4d96d1f635b39a1ce1d8b/hmailserver/source/Server/SMTP/SMTPConnection.cpp#L1762)
> 
> What you are saying this should be called right after MAIL[(*)](https://github.com/hmailserver/hmailserver/blob/e3f91404330fc5cadcc4d96d1f635b39a1ce1d8b/hmailserver/source/Server/SMTP/SMTPConnection.cpp#L426), RCPT, DATA, AUTH([*](https://github.com/hmailserver/hmailserver/blob/e3f91404330fc5cadcc4d96d1f635b39a1ce1d8b/hmailserver/source/Server/SMTP/SMTPConnection.cpp#L1783)) commands are issued when not connected through StartTLS when StartTLS is required, right?
> CheckStartTlsRequired_() was already being called for command types marked with *

Yes, it should be checked for those commands also. All commands except NOOP, EHLO, STARTTLS and QUIT should be checked with CheckStartTlsRequired_(). I can make PR also for this issue if you want.

RvdHout:
For your info, i am not the official author of hMailServer, i maintain a widely used fork with some enhancement and fixes 

CheckStartTlsRequired_() Even for HELP? and what about HELO? 

Ps,  IMAPCommandCapability.cpp in 5.6.8-Bx is OK?

If you could combine/add those checks in this PR it would be easier for @martinknafve to incorporate in 5.7.x

poundy:
> For your info, i am not the official author of hMailServer, i maintain a widely used fork with some enhancement and fixes
> 
somewhat sidetrack question though. Does your fork do releases? (or is your fork the "beta" tagged one on the website?)

OutbackMatt:
Link here >> https://www.hmailserver.com/forum/viewtopic.php?f=10&t=30193

PR #348, opened 2020-08-13
Update ListViewColumnSorter.cs

RvdHout:
Fix for sorting by DateTime in IPRanges


RvdHout:
I think i have to improve this pull, it now checks every column if it can be parsed as a DateTime value which makes it slower, i think i can do better... like @martinknafve did with Numeric values, eg: defining a Columns tag that describes it is a Integer or in this case a DateTime

PR #360, opened 2021-01-23
Update SMTPConnection.cpp

RvdHout:
Must issue STARTTLS first, for use with the pull request posted here https://github.com/hmailserver/hmailserver/pull/342

### DIFF
--- a/hmailserver/source/Server/SMTP/SMTPConnection.cpp
+++ b/hmailserver/source/Server/SMTP/SMTPConnection.cpp
@@ -605,7 +605,7 @@ namespace HM
 
       if (!current_message_) 
       {
-         EnqueueWrite_("503 must have sender first."); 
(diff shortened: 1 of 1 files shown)


martinknafve:
@RvdHout, so just so I follow - this pull request is based on #342, so I should merge #342 first right?

RvdHout:
Not based, more a completion of, see https://github.com/hmailserver/hmailserver/pull/342#issuecomment-750301859
Just for reference - RFC3207 - https://tools.ietf.org/html/rfc3207

RvdHout:
It actually makes sense once STARTTLS (Required) for some port that is can and will not accept any other 1st command as STARTTLS and if not reply with a "530 Must issue a STARTTLS command first" 

Again nothing fancy, but security scanners like OpenVAS dig this kind of stuff

PR #363, opened 2021-03-17
5.6.x & 5.7.x - Update IMAPCommandGetQuota.cpp

RvdHout:
Fix typo


RvdHout:
@martinknafve 
Any idea why the checks/build failed?

martinknafve:
@RvdHout, GitHub has removed Boost from the virtual machines used by GitHub actions, because there were few people using it and it consumed a lot of disk space. I'm leaning towards reverting to building master on TeamCity instead. Can't really use GitHub actions if they change the environment like this. The process for including OpenSSL in GitHub Actions is already very convoluted and I think using the same method for Boost would just be too time consuming.

RvdHout:
@martinknafve 
These are just some typos

Issue #239, opened 2017-10-29
oMessage.Subject and oMessage.From return nothing on OnDeliveryFailed


jacko238:
Hello
After inquiring on the forum ( https://www.hmailserver.com/forum/viewtopic.php?f=9&t=31981 ) , it seems
event OnDeliveryFailed(oMessage, sRecipient, sErrorMessage) doesn't return either oMessage.Subject nor oMessage.From though  oMessage.FromAddress works.
I use version 5.6.6-B2383
thank you.

RvdHout:
I think this only the case when triggerred by a global rule, not? 
That makes me believe the original message that should populate the oMessage properties doesn't exist anymore, eg: already delete by the rule

With 'normal' bounces, eg: a receiving email address doesn't exist on recepient server, both oMessage.Subject and oMessage.From seem populated.

RvdHout:
@martinknafve 
I'm a bit puzzled by this, sRecipientAddress is populated correctly and this value is also taken from pMessage, eg: https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/SMTP/SMTPDeliverer.cpp#L473-L481

This is strange as that would indicate pMessage is populated on L473 but no longer exist when FireOnDeliveryFailed is triggered on L481

martinknafve:
The related PR has been merged.

PR #369, opened 2021-04-12
5.6.x & 5.7.x Update SMTPDeliverer.cpp

RvdHout:
oMessage object was emtpy when called from OnDeliveryFailed in SMTPDeliverer.cpp but the documentation says the oMessage object is available in OnDeliveryFailed Event

https://github.com/hmailserver/hmailserver/issues/239

### DIFF
--- a/hmailserver/source/Server/Common/Scripting/Events.cpp
+++ b/hmailserver/source/Server/Common/Scripting/Events.cpp
@@ -84,7 +84,7 @@ namespace HM
 
          switch (pResult->GetValue())
          {
-         case 1:
+            case 1:
             {
                String sMessage;
                sMessage.Format(_T("SMTPDeliverer - Message %I64d: ")
@@ -93,7 +93,8 @@ namespace HM
 
                LOG_APPLICATION(sMessage);
 
-               PersistentMessage::DeleteObject(pMessage);             
+               //Moved to PreprocessMessage_() in SMTPDeliverer.cpp
+               //PersistentMessage::DeleteObject(pMessage);                   
 
                return false;
             }
--- a/hmailserver/source/Server/SMTP/SMTPDeliverer.cpp
+++ b/hmailserver/source/Server/SMTP/SMTPDeliverer.cpp
@@ -164,10 +164,7 @@ namespace HM
          return false;
       }
 
-      if (AWStats::GetEnabled())
-      {
-         sendersIP = MessageUtilities::GetSendersIP(pMessage);
-      }
+      sendersIP = MessageUtilities::GetSendersIP(pMessage);
 
       // Create recipient list.
       String sRecipientList = pMessage->GetRecipients()->GetCommaSeperatedRecipientList();
@@ -197,14 +194,16 @@ namespace HM
       // Run virus protection.
       if (!RunVirusProtection_(pMessage))
       {
-         LogAwstatsMessageRejected_(sendersIP, pMessage, "Message delivery cancelled during virus scanning");
+         //LogAwstatsMessageRejected_ moved to HandleInfectedMessage_()
+         //LogAwstatsMessageRejected_(sendersIP, pMessage, "Message delivery cancelled during virus scanning");
          return false;
       }
 
       // Apply rules on this message.
       if (!RunGlobalRules_(pMessage, globalRuleResult))
       {
-         LogAwstatsMessageRejected_(sendersIP, pMessage, "Message delivery cancelled during global rules");
+         //LogAwstatsMessageRejected_ moved to RunGlobalRules_()
+         //LogAwstatsMessageRejected_(sendersIP, pMessage, "Message delivery cancelled during global rules");
          return false;
       }
 
@@ -337,6 +336,8 @@ namespace HM
       if (antiVirusConfig.AVAction() == AntiVirusConfiguration::ActionDelete)
       {
          // The message should be deleted.
+         String sendersIP = MessageUtilities::GetSendersIP(pMessage);
+         LogAwstatsMessageRejected_(sendersIP, pMessage, "Message delivery cancelled during virus scanning");
 
          // Should we notify the recipient?
          if (antiVirusConfig.AVNotifyReceiver())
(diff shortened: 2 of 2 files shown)


RvdHout:
@martinknafve 
This one i am not so sure about...
yes the documentation says the OMessage object is available in OnDeliveryFailed Event, but how usefull is it?
the fix works though, it simply shifts the actual delete command till after OnDeliveryFailed

martinknafve:
This pull request moved the deletion of the message in a couple of scenarios, but not when the event `OnDeliveryStart` is triggered. When `OnDeliveryStart` is triggered, `Events::FireOnDeliveryStart` may delete the message. After that, we call `LogAwstatsMessageRejected_`. That's incorrect behavior, right?

RvdHout:
Ah...i think i missed that one (if i understand you correctly)

You mean this
```
      // Run the first event in the delivery chain
      if (!Events::FireOnDeliveryStart(pMessage))
      {
         LogAwstatsMessageRejected_(sendersIP, pMessage, "Delivery cancelled by OnDeliveryStart-event");
         return false;
      }
```


Should be like:



Events.cpp
```
   bool 
   Events::FireOnDeliveryStart(std::shared_ptr<Message> pMessage)
   {
      if (!Configuration::Instance()->GetUseScriptServer())
         return true;

      std::shared_ptr<ScriptObjectContainer> pContainer = std::shared_ptr<ScriptObjectContainer>(new ScriptObjectContainer);
      std::shared_ptr<Result> pResult = std::shared_ptr<Result>(new Result);
      pContainer->AddObject("HMAILSERVER_MESSAGE", pMessage, ScriptObject::OTMessage);
      pContainer->AddObject("Result", pResult, ScriptObject::OTResult);
      String sEventCaller = "OnDeliveryStart(HMAILSERVER_MESSAGE)";
      ScriptServer::Instance()->FireEvent(ScriptServer::EventOnDeliveryStart, sEventCaller, pContainer);

      switch (pResult->GetValue())
      {
      case 1:
         {
            String sMessage;
            sMessage.Format(_T("SMTPDeliverer - Message %I64d: ")
               _T("Message deleted. Action was taken by script subscribing to OnDeliveryStart."),
               pMessage->GetID());

            LOG_APPLICATION(sMessage);

            //Moved to PreprocessMessage_() in SMTPDeliverer.cpp
            //PersistentMessage::DeleteObject(pMessage);             

            return false;
         }
      }

      return true;
   }

```
SMTPDeliverer.cpp
```
      // Run the first event in the delivery chain
      if (!Events::FireOnDeliveryStart(pMessage))
      {
         LogAwstatsMessageRejected_(sendersIP, pMessage, "Delivery cancelled by OnDeliveryStart-event");
         PersistentMessage::DeleteObject(pMessage);
         return false;
      }
```

martinknafve:
Yeah, exactly. I can fix this. Just wanted to confirm that I was not missing something. :)

RvdHout:
Feel a bit stupid i didn't see that before :(

martinknafve:
Don't worry about it, easy mistake to make. I fixed it.

Do you agree that as a next step, it would make sense to move the `PersistentMessage::DeleteObject(pMessage);` out of the `PreprocessMessage_`-function and only call it if `PreprocessMessage_` returns false?

When `PreprocessMessage_` returns false, that means that the message delivery failed and that the message could be deleted. Right now there's 7 calls to `PersistentMessage::DeleteObject` from within `PreprocessMessage_`, but they are all followed by a `return false` so they could be replaced with:

```      
      if (!PreprocessMessage_(pMessage, sSendersIP, globalRuleResult))
      {
         // Message delivery was aborted during preprocessing.
         PersistentMessage::DeleteObject(pMessage)
         return;
      }
```

What do you think? This was actually one of the reasons I liked this pull request - the deletion can be centralized and it's easier to see that the message will be deleted when the delivery fail like this (rather than having the deletion-logic spread out in a bunch of separate classes which is hard to get an overview of)

RvdHout:
That sounds plausible and more efficient, so yes that does make sense

RvdHout:
Something i noticed....

On line 85 just before the code snippet you posted above you set: String sSendersIP; and pass it along as parameter to PreprocessMessage_() but you don't fill it's value until inside PreprocessMessage_()  (line 167)

Would it make more sense to shift that to line 85?

```
      String sSendersIP = MessageUtilities::GetSendersIP(pMessage);
      if (!PreprocessMessage_(pMessage, sSendersIP, globalRuleResult))
      {
         // Message delivery was aborted during preprocessing.
         PersistentMessage::DeleteObject(pMessage);
         return;
      }
```

martinknafve:
@RvdHout, `MessageUtilities::GetSendersIP` reads the email message file from disk to determine the address. `PreprocessMessage_`is the first place where we check that the file still exists. So if we just moved out GetSendersIP as in your example below (and run it before PreprocessMessage_), then if the file is missing then I think there would be an error, because `PersistentMessage::LoadHeader` would end up failing to load the file from disk.

I think an alternative step would be to extract the following code out of PreprocessMessage_:

```
      const String messageFileName = PersistentMessage::GetFileName(pMessage);
      if (!FileUtilities::Exists(messageFileName))
      {
         ErrorManager::Instance()->ReportError(ErrorManager::Medium, 5006, "SMTPDeliverer::DeliverMessage()", "Message " + StringParser::IntToString(pMessage->GetID()) + " could not be delivered since the data file does not exist.");
         PersistentMessage::DeleteObject(pMessage);

         return false;
      }

      if (pMessage->GetRecipients()->GetCount() == 0)
      {
         // No remaining recipients.
         ErrorManager::Instance()->ReportError(ErrorManager::Medium, 5007, "SMTPDeliverer::DeliverMessage()", "Message " + StringParser::IntToString(pMessage->GetID()) + " could not be delivered. No remaining recipients. File: " + messageFileName);

         PersistentMessage::DeleteObject(pMessage);
         return false;
      }

      sendersIP = MessageUtilities::GetSendersIP(pMessage);
```

The logic above to check for file existance and remaining recipient isn't really "preproessing" anyway - it's checking that there's anything left to do. So I think it would make sense to move it out of preprocess. Thoughts?

RvdHout:
You are the boss, i'll adapt to it :-)

Issue #371, opened 2021-04-18
MX resolve CName only


OutbackMatt:
https://www.hmailserver.com/forum/viewtopic.php?f=7&t=36270

NSLookup on same machine works, but hMailserver 5.7 b2556 won't deliver to domains with only CNAME

Some examples in thread include 'bohemia-znak.cz'

RvdHout:
Addendum

**Symptom:**  hMailServer 5.7.x doesn't resolve MX records that contain a CNAME record

https://www.rfc-editor.org/rfc/rfc2181.txt
```
10.3. MX and NS records

The domain name used as the value of a NS resource record, or part of
the value of a **MX resource record must not be an alias.** Not only is
the specification clear on this point, but using an alias in either
of these positions neither works as well as might be hoped, nor well
fulfills the ambition that may have led to this approach. This
domain name must have as its value one or more address records.
Currently those will be A records, however in the future other record
types giving addressing information may be acceptable. It can also
have other RRs, **but never a CNAME RR**.
```

According the RFC this behavior is absolutely fine, but the practical side learns us that quite a few sysadmins don't follow this RFC and do use CNAME records as MX records.
hMailServer 5.6.x resolves MX records that contain CNAME records without issues

@martinknafve 
If you agree to ignore the RFC like the 5.6.x branch does, it think 2 places need attention

DNSResolver::GetEmailServers() should call: 
GetIpAddresses(sDomainName, a_records, true) apposed to GetIpAddresses(sDomainName, a_records, false) it currently does 2x, [L188](https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp#L188) and [L224](https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp#L224)

DNSResolver::GetMXRecords() need a bit more work, eg: similar to DNSResolver::GetTXTRecords() and DNSResolver::GetTXTRecordsRecursive_()

martinknafve:
Fixed

PR #372, opened 2021-04-20
 5.7.x resolve MX records that contains CNAME record

RvdHout:
hMailServer 5.7.x doesn't resolve MX records that contain CNAME record(s)
Prevent A, AAAA and CNAME Lookup for Null MX

Details: https://github.com/hmailserver/hmailserver/issues/371

### DIFF
--- a/hmailserver/source/Server/Common/Application/Errors.txt
+++ b/hmailserver/source/Server/Common/Application/Errors.txt
@@ -1,5 +1,6 @@
 4401, Too many recursions during IP address lookup. Query: {0}
 4402, Too many recursions during TXT record lookup. Query: {0}
+4403, Too many recursions during MX record lookup. Query: {0}
 5121, PersistentACLPermission::SaveObject, ACL item of type PTAnyone not initialized correctly.
 5122, SQLCEConnection::_UpgradeDatabase(), Failed to create instance of SQL CE Engine.
 5123, SQLCEConnection::_UpgradeDatabase(), Failed to upgrade SQL CE database. HRESULT:
--- a/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp
+++ b/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp
@@ -68,7 +68,7 @@ namespace HM
    {
       if (hostName.IsEmpty())
       {
-         ErrorManager::Instance()->ReportError(ErrorManager::Medium, 5516, "DNSResolver::GetIpAddressesRecursive", "Attempted DNS lookup for empty host name.");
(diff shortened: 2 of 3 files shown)


RvdHout:
@martinknafve 
I'm not completely sure about the Null MX in https://github.com/hmailserver/hmailserver/pull/372/commits/cffd4425bb7d40e3ccbce2cbc5d162847f289110

It might be better to terminate the loop, if the MX record with lowest priority (eg: 0 for Null MX) is a Null MX ideally there won't be any other MX's

`           // Null MX, see: rfc7505 
            // https://tools.ietf.org/html/rfc7505
            if (dnsRecord.GetValue() == "." && dnsRecord.GetPreference() == 0)
            {
               dnsQueryFailure = true;
               continue;
               // or better terminate the loop?
               // break;
            }`

RvdHout:
Code below also works....
```
            // Null MX, see: rfc7505 
            // https://tools.ietf.org/html/rfc7505
            if (dnsRecord.GetValue() == "." && dnsRecord.GetPreference() == 0)
            {
               continue;
            }
```

```
"TCPIP"	10060	"2021-04-24 01:09:19.416"	"DNS MX lookup: domain.com"
"TCPIP"	10060	"2021-04-24 01:09:19.416"	"DNS - MX Result: 0 IP addresses were found."
"APPLICATION"	10060	"2021-04-24 01:09:19.416"	"SMTPDeliverer - Message 25687: No mail servers could be found for the address test@domain.com."
```

martinknafve:
Thanks :) I was thinking maybe in the "Null MX"-case, the code should do break; instead of continue;. According to the RFC, 

> A domain that advertises a null MX MUST NOT advertise any other MX RR.

.. but then of course maybe some domain owners will create a null record and then add a valid one. I'm not sure what the expected behavior would be then (the configuration would not follow the RFC). I think the current implementation of attempting to deliver if a valid host name is found is OK.

RvdHout:
Yes, exactly what i was thinking....another the RFC also states not to use CNAME in MX, but it shocking to see how mant sysadmin seem to ignore that

I think we simply must assume, when a sysadmin creates a Null MX they jnow what they are doing and should not add another MX record

martinknafve:
I added some automated tests for this to prevent it from appearing in future. To support that, I added some DNS entries for hmailserver.com:

* MX record cname-test.hmailserver.com resolves to cname.hmailserver.com which is a CNAME pointing at mail.hmailserver.com (this was not properly resolved before but now it is)
* CNAME record recursive1.hamilserver.com points at recursive2.hmailserver.com and vice versa. (this results in a logged error)

Issue #116, opened 2015-07-19
Add Return-Path before sending message to SpamAssassin


RvdHout:
SpamAssassin uses the EnvelopeFrom pseudo-header value for it's SPF checks, for some weird reason SpamAssassin is not able to find the EnvelopeFrom pseudo-header on it's own when it is passed by hmailServer, so it tries a few common values, eg:

```
envelope_sender_header Name-Of-Header
SpamAssassin will attempt to discover the address used in the 'MAIL FROM:' phase of the SMTP transaction that delivered this message, if this data has been made available by the SMTP server. This is used in the EnvelopeFrom pseudo-header, and for various rules such as SPF checking.

By default, various MTAs will use different headers, such as the following:

X-Envelope-From
Envelope-Sender
X-Sender
Return-Path

SpamAssassin will attempt to use these, if some heuristics (such as the header placement in the message, or the absence of fetchmail signatures) appear to indicate that they are safe to use. However, it may choose the wrong headers in some mailserver configurations. (More discussion of this can be found in bug 2142 and bug 4747 in the SpamAssassin BugZilla.)

To avoid this heuristic failure, the envelope_sender_header setting may be helpful. Name the header that your MTA or MDA adds to messages containing the address used at the MAIL FROM step of the SMTP transaction.

If the header in question contains < or > characters at the start and end of the email address in the right-hand side, as in the SMTP transaction, these will be stripped.

If the header is not found in a message, or if it's value does not contain an @ sign, SpamAssassin will issue a warning in the logs and fall back to its default heuristics.

(Note for MTA developers: we would prefer if the use of a single header be avoided in future, since that precludes 'downstream' spam scanning. http://wiki.apache.org/spamassassin/Env ... InReceived details a better proposal, storing the envelope sender at each hop in the Received header.)

More info:
https://www.hmailserver.com/forum/viewtopic.php?f=8&p=177529
```

RvdHout:
Is there some sort of notification to see if a bug/issue is looked into? 

The fact the Return-Path is missing in messages passed to spamassassin is realy i bug in my opinion as it was part of your code in the 4.x versions, see: https://www.hmailserver.com/forum/viewtopic.php?p=16474#p16474

martinknafve:
Hi

This is something i plan on looking into for 5.7.

Martin

RvdHout:
Oke, thanks

RvdHout:
[...]

PR #374, opened 2021-04-24
5.7.0 SpamAssassin X-hMailServer-Envelope-From

RvdHout:
Use X-hMailServer-Envelope-From header in favour of first adding and later removing the Return-Path Header for use with SpamAssassin

This basically undo's this pull
https://github.com/hmailserver/hmailserver/issues/116

### DIFF
--- a/hmailserver/source/Server/Common/AntiSpam/SpamTestSpamAssassin.cpp
+++ b/hmailserver/source/Server/Common/AntiSpam/SpamTestSpamAssassin.cpp
@@ -56,27 +56,21 @@ namespace HM
       std::shared_ptr<Message> pMessage = pTestData->GetMessageData()->GetMessage();
       const String sFilename = PersistentMessage::GetFileName(pMessage);
 
-      String sEnvelopeFrom = pTestData->GetEnvelopeFrom();
+      // SMTP servers making final delivery MAY/SHOULD remove Return-path header fields before adding their own. See: rfc2821 and rfc5321
+      if (!pTestData->GetMessageData()->GetReturnPath().IsEmpty())
+      {
+         do {
+            pTestData->GetMessageData()->DeleteField("Return-Path");
+         } while (!pTestData->GetMessageData()->GetReturnPath().IsEmpty());
+      }
 
-      // Add "X-hMailServer-Envelope-From" header
-      // Add following to SpamAssassin local.cf: envelope_sender_header X-hMailServer-Envelope-From
+      // Add Return-Path as topmost header to help SpamAssassin with its SPF checks.
+      // SpamAssassin default rules and custom rules also rely on Return-Path header being present
+      // We delete this header again after SpamAssassin checking has completed
       std::vector<std::pair<AnsiString, AnsiString>> fieldsToWrite;
-      fieldsToWrite.push_back(std::make_pair("X-hMailServer-Envelope-From", sEnvelopeFrom));
-
+      fieldsToWrite.push_back(std::make_pair("Return-Path", pTestData->GetEnvelopeFrom()));
       TraceHeaderWriter writer;
       writer.Write(sFilename, pMessage, fieldsToWrite);
-
-      // Add Return-Path as topmost header if none exist (ExternalAccount download?)
-      // For SpamAssassin default rules and custom rules that rely on Return-Path header being present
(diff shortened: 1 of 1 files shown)


RvdHout:
On second thought I am not so sure...
I checked the default and some custom SA rulesets i use and there are quite a few rules that depend on the Return-Path header

I know I could override those rules to use the Envelope-Sender header instead, but maybe not the best solution for novice users, Return-Path is more globally accepted too...I think

On the other hand, first adding and later removing the Return-Path header is a bit messy...

@martinknafve 
Thoughts?

RvdHout:
If figured maybe we should just insert both headers, eg:

Return-Path as topmost header (if none exists) before sending to SA (and delete this one afterwards)
We use this for SpamAssassin default rules and custom rules that rely on Return-Path header being present

------------------------------------------

X-hMailServer-Envelope-From as additional header, i seem to have thought about something like that back in [2015](https://www.hmailserver.com/forum/viewtopic.php?p=177520#p177520) already (this one stays in the message)

One could utilize this header by adding the following to SpamAssassin local.cf: envelope_sender_header X-hMailServer-Envelope-From

RvdHout:
@martinknafve 

If figured maybe we should just insert both headers, eg:

Return-Path as topmost header (if none exists) before sending to SA (and delete this one afterwards)
We use this for SpamAssassin default rules and custom rules that rely on Return-Path header being present

------------------------------------------

X-hMailServer-Envelope-From as additional header, i seem to have thought about something like that back in [2015](https://www.hmailserver.com/forum/viewtopic.php?p=177520#p177520) already (this one stays in the message)

One could utilize this header by adding the following to SpamAssassin local.cf: envelope_sender_header X-hMailServer-Envelope-From

RvdHout:
@martinknafve 

Other possibility would be to delete any existing Return-Path headers before were are adding ours, but I'm not sure if we are allowed to? In that scenario we can solely depend on the Return-Path header and no customization to local.cf  in SpamAssassin is needed

RvdHout:
@martinknafve, can you explain me why there a 2 different types of Return-Path values used in hmailserver?

**In MessageData.cpp**
SetFieldValue("Return-Path", "<" + sReturnPath + ">"); 

**In LocalDelivery.cpp**
fieldsToWrite.push_back(std::make_pair("Return-Path", pMessage->GetFromAddress()));

Why the first one contains a pair of angle brackets and the second not? Shouldn't it be better to generalize the format?

**The use of angle brackets is mentioned in:**
https://tools.ietf.org/html/rfc2822  (3.6.7. Trace fields)
https://tools.ietf.org/html/rfc822

**But those do not mention the use of angle brackets** 
https://tools.ietf.org/html/rfc2821
https://tools.ietf.org/html/rfc5321

Both mention a null Return-Path, should be in the format: Return-Path: <>

---------------------------------------------------

**These RFC's mention we may delete any existing Return-Path header and add our own**

https://tools.ietf.org/html/rfc2821
https://tools.ietf.org/html/rfc5321

> It is sometimes difficult for an SMTP server to determine whether or
>    not it is making final delivery since forwarding or other operations
>    may occur after the message is accepted for delivery.  Consequently,
>    any further (forwarding, gateway, or relay) systems MAY remove the
>    return path and rebuild the MAIL command as needed to ensure that
>    exactly one such line appears in a delivered message.
> 
>    A message-originating SMTP system SHOULD NOT send a message that
>    already contains a Return-path header field.  SMTP servers performing
>    a relay function MUST NOT inspect the message data, and especially
>    not to the extent needed to determine if Return-path header fields
>    are present.  SMTP servers making final delivery **MAY remove** Return-
>    path header fields before adding their own.

Neustradamus:
@RvdHout: Why have you closed this PR?

RvdHout:
@Neustradamus 
Because it was old, and no good

Neustradamus:
@RvdHout: Ok all your closed PRs have been replaced by a better one?

RvdHout:
Why do you care? What are your intentions here, just complain and ask questions?

Neustradamus:
The goal is to do not forget good contributions :)

RvdHout:
Right, and this one was closed, so don't be ignorant and ignore the fact i closed this myself

PR #377, opened 2021-04-26
5.7.0 InstallLocation RegKey 64-bit

RvdHout:
On 64-bit, this registry key is created in HKLM\SOFTWARE\WOW6432Node\hMailServer\InstallLocation which is the 32-bit location
I believe on 64-bit this has to be HKLM\SOFTWARE\hMailServer\InstallLocation

### DIFF
--- a/hmailserver/installation/hMailServerInnoExtension.iss
+++ b/hmailserver/installation/hMailServerInnoExtension.iss
@@ -848,7 +848,7 @@ begin
 	begin
    // Create a registry key that tell
 	  // other apps where we're installed.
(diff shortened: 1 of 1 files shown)


RvdHout:
Bump (@martinknafve i noticed you commited some installer fix, this one in my opinion should be included as well)

martinknafve:
Does this cause a problem? The reason I kept the old one was compatibility with the previous version. If you install 5.6.x first, then it's put into the x86-location. If you later on upgrade to 5.7, then it keeps using the old ones. If I recall correctly, it was needed for uninstalls to work properly if you installed 5.6.x (32bit) then 5.7 (64bit) and then uninstalled hMailServer via Programs & features.

PR #383, opened 2021-06-19
DKIM signature for domain aliases

mstricek:
Added option to allow DKIM signing for domain aliases using key and selector from the parent domain.
As requested https://www.hmailserver.com/forum/viewtopic.php?f=2&t=29029&p=208275
(main reason was that I needed it myself)

### DIFF
--- a/hmailserver/source/Server/COM/InterfaceDomain.cpp
+++ b/hmailserver/source/Server/COM/InterfaceDomain.cpp
@@ -1198,6 +1198,47 @@ STDMETHODIMP InterfaceDomain::put_DKIMSignEnabled(VARIANT_BOOL newVal)
    }
 }
 
+STDMETHODIMP InterfaceDomain::get_DKIMSignAliasesEnabled(VARIANT_BOOL* pVal)
+{
+   try
+   {
+      if (!object_)
+         return GetAccessDenied();
+
+      if (!authentication_->GetIsDomainAdmin())
+         return authentication_->GetAccessDenied();
+
+      *pVal = object_->GetDKIMAliasesEnabled() ? VARIANT_TRUE : VARIANT_FALSE;
+
+      return S_OK;
+   }
+   catch (...)
+   {
+      return COMError::GenerateGenericMessage();
+   }
+}
+
+STDMETHODIMP InterfaceDomain::put_DKIMSignAliasesEnabled(VARIANT_BOOL newVal)
+{
+   try
+   {
+      if (!object_)
+         return GetAccessDenied();
+
+      if (!authentication_->GetIsDomainAdmin())
+         return authentication_->GetAccessDenied();
+
+      object_->SetDKIMAliasesEnabled(newVal == VARIANT_TRUE);
+
+      return S_OK;
+   }
+   catch (...)
+   {
+      return COMError::GenerateGenericMessage();
+   }
+}
+
+
 STDMETHODIMP InterfaceDomain::get_DKIMSelector(BSTR *pVal)
 {
    try
--- a/hmailserver/source/Server/COM/InterfaceDomain.h
+++ b/hmailserver/source/Server/COM/InterfaceDomain.h
@@ -151,6 +151,8 @@ END_COM_MAP()
    STDMETHOD(put_DKIMBodyCanonicalizationMethod)(/*[in]*/ eDKIMCanonicalizationMethod newVal);
    STDMETHOD(get_DKIMSigningAlgorithm)(/*[out, retval]*/ eDKIMAlgorithm *pVal);
    STDMETHOD(put_DKIMSigningAlgorithm)(/*[in]*/ eDKIMAlgorithm newVal);
+   STDMETHOD(get_DKIMSignAliasesEnabled)(/*[out, retval]*/ VARIANT_BOOL* pVal);
+   STDMETHOD(put_DKIMSignAliasesEnabled)(/*[in]*/ VARIANT_BOOL newVal);
 
    // dkim end.
 private:
--- a/hmailserver/source/Server/Common/AntiSpam/DKIM/DKIMSigner.cpp
+++ b/hmailserver/source/Server/Common/AntiSpam/DKIM/DKIMSigner.cpp
@@ -10,8 +10,11 @@
 
 #include "../../BO/Message.h"
 #include "../../BO/Domain.h"
+#include "../../BO/Alias.h"
+#include "../../Application/ObjectCache.h"
 #include "../../Cache/CacheContainer.h"
 #include "../../Util/Hashing/HashCreator.h"
+#include "../../BO/DomainAliases.h"
 
 #ifdef _DEBUG
 #define DEBUG_NEW new(_NORMAL_BLOCK, __FILE__, __LINE__)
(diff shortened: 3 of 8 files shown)


RvdHout:
I like it, thx! 👍

RvdHout:
One thing i notice when i send a few test mails to check-auth@verifier.port25.com

Using the main domain it reports:

```
DomainKeys check details:
----------------------------------------------------------
Result:         pass

```
Using the alias it reports:

```
----------------------------------------------------------
DKIM check details:
----------------------------------------------------------
Result:         pass (signature verifies; identity doesn't match any headers)
```

Not sure if that is a problem (or might be a problem if future) but it seems to work OK (both spamassassin as hmailserver DKIM check verify the alias signed mail correctly

mstricek:
Interesting. Inspired by your post I have checked my settings against the check-auth@verifier.port25.com and in both cases I got pass including the match (domains and names anonymized):
```
----------------------------------------------------------
DKIM check details:
----------------------------------------------------------
Result:         pass (matches From: somename@domain1.eu)
ID(s) verified: header.d=domain1.eu
```
and the alias then
```
----------------------------------------------------------
DKIM check details:
----------------------------------------------------------
Result:         pass (matches From: othername@domain2.eu)
ID(s) verified: header.d=domain2.eu
```
But you have "domain.nl" in both cases - in the second case the header.d should be the alternate domain.

RvdHout:
@mstricek 
That is strange! 
Did you create a additional DKIM DNS record for domain2.eu? Or just the one for domain1.eu?

mstricek:
I have independent DKIM DNS records for each domain. Basically on the DNS side the domains are separated completely as it would be for independent domains in hMailServer, they just use the same DKIM keys.

RvdHout:
@mstricek 
I only have a DKIM DNS record for the main domain.nl, the alias domain.com has none (but always good to know this works/verifies as well 😀)

Independent DKIM DNS records you say, but 100% identical I suppose or if they are different did you do something fancy to share the key between them?

RvdHout:
I don't get what you said in your last message, you sign the message using the key and selector of the main domain, right?

So the DKIM header in the message received will always read:
```
dkim-signature: v=1; a=rsa-sha256; d=domain.nl; s=mail; c=relaxed/relaxed; q=dns/txt;
 h=From:Subject:Date:Message-ID:To:MIME-Version:Content-Type; bh=***BASE64STRING***
```
Not? Or did i miss something?

RvdHout:
Looks like i missed a part in your commit, https://github.com/hmailserver/hmailserver/commit/5865658693140ec68c95a303fc58207ff96a2289#diff-6a1e94607eb96f51a90716a96729748ef40d22bb56ca5d49c3127fc7d5633fa5R59

That should explain why it still sign with d=domain.nl in the header

But now we know this works/verifies too, is it still worth to rewrite d=domain.nl to domain.com in header? And require 2 independent DKIM DNS records instead a single one?

RvdHout:
Maybe make it optional?
Eg: another checkbox, require Independent DNS records

mstricek:
I think the idea is to match the domain from the sender address with the domain in the header and then check the signature using the key from the DKIM DNS record for that domain. So rewriting the domain in the header will be always treated as a "half baked" solution in the best case, in a worst case as a spoofing attempt. 

If someone is sending emails from a certain domain then I would expect they have control over the DNS records for that domain - they must have been able to setup MX record pointing to their email server. So adding the DKIM, SPF and DMARC records should be available to them as well.

RvdHout:
You are rewriting the domain in the header yourself, so you say you created a "half baked" solution? 😉 (Yeah i understand you meant without rewriting domain the header)

I (accidentally) found out there is no requirement for the domain header to match the From header, you still need to know the selector to use/abuse it.

RvdHout:
A pretty big company here in the Netherlands sends out their promotional mails the same way i accidentally did, eg: the From address domain reads company.nl but it dkim signed using company.be domain in the dkim header.

So as it is no requirement I would recommend to make it optional

mstricek:
You are correct, the signing domain does not need to have relation to the sender email address domain. So the signing for aliases could have been enabled by default.

It seems like I have overengineered things a bit :-)

Since I have the checkbox there already I just can change its meaning to "Use alias DKIM domain" or something like that and keep the signing checkbox as a master switch. When checked the main domain as well as the aliases get signed.

mstricek:
One more thing - I have retraced my thought process to find out why I implemented like that - I was looking for it for myself and found out the feature request, then went to the code and saw how it was done there and from that I made a conclusion that the sender domain need to be retained in DKIM header. I treated the existing code as a correct implementation and thus I wanted to add the option to sign aliases while retaining compatibility with the original implementation - this is why I have added the checkbox.

Today, after our productive discussion, I would treat the missing signing for aliases in the original code as a bug and simply add the signing for them without implementing any checkboxes - only the DKIMSigner.cpp would change in this case and nothing else :-)

RvdHout:
@mstricek 
I understand what you are saying, but....
There are always people that want it to be perfect, eg: rewiting -d parameter in the DKIM e-mail header to match the From address

@martinknafve 
Any input on this?

RvdHout:
@mstricek, here is a interestiing read on DKIM alignment 
https://mxtoolbox.com/dmarc/dkim/dkim-alignment

RvdHout:
Mmmm, very strange i included the above adkim=s parameter to my DMARC DNS Record but it still says it passes fine using my old built that did not rewrite the -d DKIM header....clueless 🤔

Authentication-Results: mx.google.com
```
       dkim=pass header.i=@domain.nl header.s=dkim header.b=muT9lRb5;
       spf=pass (google.com: domain of ruud@domain.com designates xxx.xxx.xxx.xxx as permitted sender) smtp.mailfrom=ruud@domain.com;
       dmarc=pass (p=QUARANTINE sp=REJECT dis=NONE) header.from=domain.com
```

mstricek:
> https://mxtoolbox.com/dmarc/dkim/dkim-alignment
> 
It reminds me that I probably read this one when I was looking for an explanation of the difference between "Simple" and "Relaxed" Header and body method radio buttons in the hMailServer DKIM Signing tab.

In fact this may be the factor which lead me to decide selecting the signing domain based on the sender address.

RvdHout:
It very odd outlook, google and yahoo do not seem to care my DKIM record is not aligned with message.From address, not?

**Another useful testing tool**
https://dmarcian.com/dmarc-inspector/

You can also run a quick DMARC check by sending an email with any subject/content to check@dmarcly.com. After DMARCLY receives the email, it will perform an SPF/DKIM/DMARC authentication analysis on the sender domain and send the results back to you as a report.

RvdHout:
Could it be I am misreading/not understanding what is written by Google:

> To pass DMARC, a message must pass at least one of these checks:
> 
> 

- SPF authentication and SPF alignment

> 

- DKIM authentication and DKIM alignment

> A message fails the DMARC check if the message fails both:
> 
> 

- SPF (or SPF alignment)

> 

- DKIM (or DKIM alignment)

https://support.google.com/a/answer/10032169#alignment

RvdHout:
Yes, it seems google's explanation of DMARC is absolutely correct, message will pass DMARC check if SPF authentication and SPF alignment **or** DKIM authentication and DKIM alignment passes the checks.

The DKIM defined policy check will fail using adkim=s parameter in DMARC DNS Record but the DKIM signature still validates alright, so you have to question how valuable the adkim=s parameter in DMARC record is.... that's probably explains why relaxed is the default parameter for adkim parameter

RvdHout:
Disable checkbox for domains without domain aliases, in ucDomain.cs

`checkDKIMSignAliasesEnabled.Enabled = _domain.DomainAliases.Count > 0;`

And maybe also add an additional check while saving

`_domain.DKIMSignAliasesEnabled = checkDKIMSignAliasesEnabled.Checked && _domain.DomainAliases.Count > 0;`

martinknafve:
Maybe I misinterpret the above, but it looks like this is not ready to be merged, correct?

mstricek:
Well, it depends what features the users expect. The implementation follows the recommendations and during the past year and half I am using it I have got no problem with it.

Looking at the last comment it is just a cosmetic change in the UI which does not affect functionality.

martinknafve:
Do you agree this could be merged as-is, @RvdHout ?

RvdHout:
In "my" build i have [these](https://github.com/hmailserver/hmailserver/pull/383#issuecomment-905340890) additional checks in gui

Basically it only allows you to enable this setting only if domain alias exists....it works as is, but no reason to enable this option if no domain aliases exist (in my opinion)

mstricek:
@RvdHout How will it behave in the following scenario:
1) domain aliases exist
2) enable aliases signing
3) remove domain aliases
save settings, then
4) add domain aliases
save settings

Will the aliases signing be still enabled? Or it got disabled due to the step #3 without user noticing?

martinknafve:
For this to be merged, WebAdmin needs to be updated (not just hMailServer Administrator) and tests needs to be added.

Neustradamus:
@mstricek: Any progress on your PR?

Issue #299, opened 2019-08-19
AUTH PLAIN not working with libcurl


tomy2105:
I'm using 3rd party program which uses libcurl to send mail to hMailserver using AUTH PLAIN (which is in new libcurl preferred over AUTH LOGIN) and it does not work.

Looking at the relevent RFC (https://tools.ietf.org/html/rfc4616) it seems that libcurl is correctly forming base64 encoded usernameNULusernameNULpassword sequence.

However looking at https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/SMTP/SMTPConnection.cpp method AuthenticateUsingPLAIN_(const String &sLine) I see code is looking for tab delimiter and not nul.

Moreover, only looking briefly at the code, not sure if code handles double username correctly either.

Could someone confirm if I'm correct to say hMailserver AUTH PLAIN has problems?

Thanks in advance.

martinknafve:
This was fixed as part of this:
https://github.com/hmailserver/hmailserver/pull/385

RvdHout:
@martinknafve
In my opinion and to my knowledge hmailservers accepts both /0 and /t as delimiters when using AUTH PLAIN

```
\0: \0username@domain.com\0P@ssword
\0: AHVzZXJuYW1lQGRvbWFpbi5jb20AUEBzc3dvcmQ=


220 Welcome to mail.domain.nl
EHLO mailserver.com
250-mail.domain.nl
250-SIZE 52428800
250-STARTTLS
250-AUTH LOGIN PLAIN
250 HELP
AUTH PLAIN AHVzZXJuYW1lQGRvbWFpbi5jb20AUEBzc3dvcmQ=
235 authenticated.
QUIT
221 goodbye
```

```
\t: \tusername@domain.com\tP@ssword
\t: CXVzZXJuYW1lQGRvbWFpbi5jb20JUEBzc3dvcmQ=


220 Welcome to mail.domain.nl
EHLO mailserver.com
250-mail.domain.nl
250-SIZE 52428800
250-STARTTLS
250-AUTH LOGIN PLAIN
250 HELP
AUTH PLAIN CXVzZXJuYW1lQGRvbWFpbi5jb20JUEBzc3dvcmQ=
235 authenticated.
QUIT
221 goodbye
```

in code we split them using /t, see:
https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/SMTP/SMTPConnection.cpp#L257
[...]

PR #385, opened 2021-07-21
SASL deliminator from \t to \0 ref RFC4616

large:
RFC4616 describe the deliminator to be NUL not TAB

From chapter 2, ref https://tools.ietf.org/search/rfc4616:
"The authorization identity (authzid), authentication identity
(authcid), password (passwd), and NUL character deliminators SHALL be
transferred as [UTF-8] encoded strings of [Unicode] characters. As
the NUL (U+0000) character is used as a deliminator, the NUL (U+0000)
character MUST NOT appear in authzid, authcid, or passwd productions"

Test server with GSASL: http://ftp.gnu.org/gnu/gsasl/

Did not understand what happened before I saw the code, see forum thread
https://www.hmailserver.com/forum/viewtopic.php?f=7&t=37063

### DIFF
--- a/hmailserver/source/Server/IMAP/IMAPCommandAuthenticate.cpp
+++ b/hmailserver/source/Server/IMAP/IMAPCommandAuthenticate.cpp
@@ -63,7 +63,7 @@ namespace HM
 
 		sParam = pParser->GetParamValue(pArgument, 1);
 		StringParser::Base64Decode(sParam, sDecode64);
-		std::vector<String> plain_args = StringParser::SplitString(sDecode64, "\t");
+		std::vector<String> plain_args = StringParser::SplitString(sDecode64, "\0");
 
 		if (plain_args.size() != 3)
 			return IMAPResult(IMAPResult::ResultBad, "Command has malformed base64 token.");


RvdHout:
@martinknafve see:
https://github.com/hmailserver/hmailserver/issues/299#issuecomment-1473474857

Thoughts?

large:
Well, if the standard is still valid, the use of "SHALL" indicates that tab is not valid, NUL **SHALL** be used.

https://www.rfc-editor.org/rfc/rfc2119#section-1
"1. MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
   definition is an absolute requirement of the specification"

But if there are differences from old system, why not support them both?

RvdHout:
Could be it is needed for SASL but not for AUTH PLAIN for SMTP, i find it remarkable you claim it is not working when it seems to be working fine for SMTP using both /0 and /t

large:
When I tested this 2 years ago, it did not work. What happens between here is out of my control.

My PR was a fix to comply with standards. If hMailserver devs want to extend it, they are free todo so 😁

RvdHout:
@martinknafve 
People are experiencing high memory usage on 5.7 build 2630, someone [pinpointed the issue](https://www.hmailserver.com/forum/viewtopic.php?p=246429#p246429) back to this commit
I made him a test build where i only reverted the delimiter, eg: changed \0 back to \t and he says the problem went away, strange, not?

Full conversation
https://www.hmailserver.com/forum/viewtopic.php?f=7&t=40305&p=246442#p246442

martinknafve:
@RvdHout, not sure, could very well be related. I reverted this PR. If it's caused by this I'll troubleshoot later (I have a suspicion what it is).

RvdHout:
@martinknafve, find it a bit strange that some of us hadn't had this issue at all and other did even when SASL was disabled

PR #396, opened 2021-10-04
5.7.0-DNSServer INI Setting(s)

RvdHout:
_Boolean_ **UseDNSCache** (not mine, but something Søren Raagaard Rathje created to bypass use of systems dnscache)

_String_ **DNSServer**
I have a local BIND9 caching en forwarding server, unfortunately Hmailserver by default uses system's default DNS 
This setting can be used to supply a single (local) ip4 IP-Address to force Hmailserver to use that DNS server without having to change system default DNS settings

Eg: in my case both hmailserver as spamassassin are configured to use the same local (ubuntu VM) Bind9 DNS caching en forwarding server instance, this way repeating lookups are cached and forwarding rules are shared, there are a few DNSBL's that require this.

```
/* Disable forwarding for DNSBL queries */

zone "multi.uribl.com" { type forward; forward first; forwarders {}; };
zone "multi.surbl.org" { type forward; forward first; forwarders {}; };
zone "list.dnswl.org" { type forward; forward first; forwarders {}; };
zone "all.spamrats.com" { type forward; forward first; forwarders {}; };
zone "grey.uribl.com" { type forward; forward first; forwarders {}; };
zone "black.uribl.com" { type forward; forward first; forwarders {}; };
```

### DIFF
--- a/hmailserver/source/Server/Common/Application/IniFileSettings.cpp
+++ b/hmailserver/source/Server/Common/Application/IniFileSettings.cpp
@@ -59,7 +59,8 @@ namespace HM
       blocked_iphold_seconds_(0),
       smtpdmax_size_drop_(0),
       backup_messages_dbonly_(false),
-      add_xauth_user_ip_(false)
+      add_xauth_user_ip_(false),
+      use_dns_cache_(true)
       
    {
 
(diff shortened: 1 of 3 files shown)


martinknafve:
Merged. Thanks. :)

PR #399, opened 2021-11-12
Added integer overflow check for SPF macro segment count

nathaniel-bennett:
The current version of RMSPF.cpp allows for an arbitrary number of digits to be read into the SPF macro segment count; as a result, an SPF record with a macro such as %{d2147483648} would lead to integer overflow in the `num` variable. While this does not lead to any unexpected or dangerous behavior in the current version of the source, an added check would make sure that no future errors come up because of the overflow.

As well (just in case it's needed), I give full permission to license the modifications I've made to the source code under whatever license you use, with or without any attribution.


martinknafve:
Thanks!

PR #415, opened 2022-01-24
5.7.0 (Optionally) Prefer IPv6 over IPv4

RvdHout:
If IPv6 is available and the option 'Prefer IPv6 over IPv4' is checked then IPv6 DNS lookups are performed before IPv4 DNS lookups

PreferIPv6


martinknafve:
I merged the PTR pull request which changed db version from 5700 to 5702. This changes it from 5700 to 5701. I'm not sure how to resolve that conflict. I could update this one so that instead changes version from 5702 to 5703. Because if I merge this one as is, there will be one script changing version from 5700 to 5701 and another changing from 5700 to 5702, which won't work.

Any thoughts on that?

RvdHout:
Ah yes i used the master branch as starting point for both pull request, not so clever.... i can change these so it reads 5703, OK?

RvdHout:
@martinknafve 
Would you want the database (upgrade) scripts to go from 5700 -> 5703 in one go? Or from 5702 to 5703?

martinknafve:
Please change so it goes from 5702 to 5703. _master_ is already on 5702 now.

PR #428, opened 2022-03-29
5.7.0 mime recipient headers

RvdHout:
Add MIME Recipient Headers as configurable comma delimited string of headers defined in External download properties (stored in database) and we parse those if "Deliver to recipients in MIME headers" is enabled.

Gives better flexibility on to which headers to parse, sometimes To and/or CC are desired to be left out, and additional non-rfc headers can be added to this array easily, for example: X-Original-To

160633347-97a21b03-c183-4ffe-a445-12b64a6b0f84

Base: https://github.com/hmailserver/hmailserver/pull/415

This fixes and gives options 2 these questions/issues:
https://github.com/hmailserver/hmailserver/issues/250
https://www.hmailserver.com/forum/viewtopic.php?f=2&t=38418


martinknafve:
The following commits seems unrelated to the feature you're adding, right?

* [(Optionally) Prefer IPv6 over IPv4](https://github.com/hmailserver/hmailserver/pull/428/commits/1b0e78b22e8e93e8d2f618ebe2dcb70cbd9c8c10) should not have been part of this pull request, right?

* [Merge branch 'master' into 5.7.0-Optionally-Prefer-IPv6-over-IPv4](https://github.com/hmailserver/hmailserver/pull/428/commits/f4db5850b08b00ddb89f5f220bc20e5e8fb2bc9a) also seems unrelated.

* [database 5703](https://github.com/hmailserver/hmailserver/pull/428/commits/d123fc4641358ebab7ac1768d07d003490ef4c55)

And maybe more (I didn't check details of all). It is not really possible to just merge this pull request, because large parts of it seems to be unrelated to the change you're adding.

RvdHout:
Last time you asked if i could incrementally add the database changes, eg: 5700, 5702
[Prefer IPv6 over IPv4 (https://github.com/hmailserver/hmailserver/pull/428/commits/1b0e78b22e8e93e8d2f618ebe2dcb70cbd9c8c10) pushes database to 5703 and [](https://github.com/RvdHout)@RvdHout
and this [Specify MIME Recipient Headers](https://github.com/hmailserver/hmailserver/pull/428/commits/2504588ac37ddbe45b8c33114c474ac1e142324d) to 5704 

so yes this branch depends on [Prefer IPv6 over IPv4 (https://github.com/hmailserver/hmailserver/pull/428/commits/1b0e78b22e8e93e8d2f618ebe2dcb70cbd9c8c10) so i think github added this automatically... i think if you first commit that one, this one will become more clear

martinknafve:
It looks like this replaces the following PR, is that correct?
https://github.com/hmailserver/hmailserver/pull/251

(Maybe "replaces" is the wrong word, but it looks like the 2 PR:s attempt to solve same issue in different ways)

RvdHout:
Yes, instead of a fixed list of possible values this gives the "Admin" the opportunity to specify exactly what MIME Recipient Headers are allowed and should be respected, there are so many possibilities, most none official RFC, this fallbacks to earlier fixed list of values

martinknafve:
There are conflicts in 5 files in this PR. Is it possible for you to resolve them?

martinknafve:
The SQL scripts for SQL Server, SQL Server Compact Edition and PostgreSQL contained syntax errors.

Fixed here:

* https://github.com/hmailserver/hmailserver/commit/6c20f2c13c9c87aa1c5849b79389a57d36daccd6
* https://github.com/hmailserver/hmailserver/commit/f4be7506468ea91942bfd5a144a7d5a56b037882

OutbackMatt:
I got an error when this script updated for MySQL

COLUMN `famimerecipientheaders` already exists.
I had to remove the first line of this SQL statement, ie remove the line below to get the DB Updater to complete.

"ALTER TABLE `hm_fetchaccounts` ADD COLUMN `famimerecipientheaders` varchar(255) NOT NULL DEFAULT 'To,CC,X-RCPT-TO,X-Envelope-To';"

I had run one version of @RvdHout 's build from the forum for 5.7 a little while ago, perhaps that was set then.

No real issue - just reporting in

I went from version B2632 to B2639 this morning

RvdHout:
@OutbackMatt 
That is simply impossible as this update script is for dbversion 5704, mine uses 5705 so this update never should be triggered (unless of course you didn't follow the instructions that came with that build in the first place)

OutbackMatt:
The version of your that I downloaded was 2604.17 And that was months ago.
I've been back to using the TeamCity builds since 13th March.

RvdHout:
Und? So you reverted without reverting the database changes, eg: made a mess of it!
What would you expect would happen?

OutbackMatt:
OK, no problem.

Just commented today as this is first issue that I've had.
It's all good. Love your work.

Issue #420, opened 2022-02-08
ClamAV via TCP is not working after Update to 0.104


LordDragonus:
If using ClamAV via TCP with version 0.104 or higher, hMailServer shows Message Box with „UNKNOWN COMMAND“.

This is because hMailServer is using the command "STREAM" as shown here https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/AntiVirus/ClamAVVirusScanner.cpp#L64
It was deprecated and has been deleted now. The right command should be "INSTREAM" now, as shown in the manpage: https://manpages.debian.org/testing/clamav-daemon/clamd.8.en.html

Forum thread: https://www.hmailserver.com/forum/viewtopic.php?f=7&t=37507

RvdHout:
Hello @LordDragonus 

When was this deprecated? With version 0.103.x? 
The change should be fairly easy, but what to do with backward compatibility? Not everyone is (able to) use/using the latest version of ClamAV, many of us are still on 0.103.x or even older versions

RvdHout:
I did some test, i can send the data to clamav using a single socket connection and inquiry the INSTREAM command but have trouble receiving data back using SynchronousConnection::ReadUntil(const AnsiString &delimiter, AnsiString &readData)

Not sure what i am missing here...

SorenRR:
@RvdH you wrote earlier that you could send size and data but not receive reply... Well, I have trouble converting to 4 byte big endian (probably b.c. I haven't "broken the code yet") BUT... In 5.6.8 I made these changes and I DO read the reply from ClamAV...  Join forces?

https://github.com/SorenRR/hmailserver/commit/4deff1500c3f1f10273b570185e06c172bb97e24

RvdHout:
@SorenRR 

I think you/we are completely on the wrong track, build a example project using the C# code posted at the forums and see the obvious difference with what we are doing

```
00000000  6E 49 4E 53 54 52 45 41  4D 0A 00 00 00 44 58 35   nINSTREA M....DX5 
00000010  4F 21 50 25 40 41 50 5B  34 5C 50 5A 58 35 34 28   O!P%@AP[ 4\PZX54( 
00000020  50 5E 29 37 43 43 29 37  7D 24 45 49 43 41 52 2D   P^)7CC)7 }$EICAR- 
00000030  53 54 41 4E 44 41 52 44  2D 41 4E 54 49 56 49 52   STANDARD -ANTIVIR 
[...]

PR #433, opened 2022-06-16
5.7.0-ClamAV-INSTREAM

RvdHout:
Something that was laying on the shelf but never made it into a pull request until now, most of the credits go to Søren Raagaard Rathje, see https://github.com/hmailserver/hmailserver/issues/420

Any version newer than ClamAV 0.103.x, eg: ClamAV 0.104.x and 0.105.x require INSTREAM command and support for the previously used STREAM command is dropped.

Actually the STREAM command is "deprecated" since 2009 (but continued to work up to 0.103.x)

Since the INSTREAM command has been added 2009 too, there should be no problem for backward compatibility. eg: All ClamAV versions since 0.95 should work with INSTREAM.

### DIFF
--- a/hmailserver/source/Server/Common/AntiVirus/ClamAVVirusScanner.cpp
+++ b/hmailserver/source/Server/Common/AntiVirus/ClamAVVirusScanner.cpp
@@ -48,41 +48,24 @@ namespace HM
    VirusScanningResult
    ClamAVVirusScanner::Scan(const String &hostName, int primaryPort, const String &sFilename)
    {
-      LOG_DEBUG("Connecting to ClamAV virus scanner...");
-
-      int streamPort = 0;
+      union
+      {
+         uint32_t integer;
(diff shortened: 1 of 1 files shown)


RvdHout:
Additional resources:
https://www.hmailserver.com/forum/viewtopic.php?f=7&t=37507
https://github.com/hmailserver/hmailserver/issues/420

RvdHout:
@martinknafve 

Not sure if this could be coded more efficient, and what is the best approach, eg: memcpy() vs union as discussed here: https://github.com/hmailserver/hmailserver/issues/420

Quite a few people 'waiting' for this, so if you could somehow hurry up with this that would be great 😜

palinkas-jo-reggelt:
@martinknafve

> Quite a few people on the forums requested this and are 'waiting' for this, so if you could somehow hurry up with this that would be great 😜

Yes, please.

Issue #167, opened 2016-05-02
HM5157.  Hmailserver winsock 2. There was a communication error with SpamAssassin


jimimaseye:
As logged and discussed many times.  Latest discussion with author of Spamassassin (Harald 'Binkle') ruling out Spamassassin as the cause.  Further high level analysis suggesting the cause (code or 'Boost' library), with agreement to look in to, found from here:  https://www.hmailserver.com/forum/viewtopic.php?p=181600#p181600

Boost diagnosis from here: https://www.hmailserver.com/forum/viewtopic.php?p=184625#p184625

>  I have discovered through my proxy program that I, too, occasionally get the socket close notification (or EOF as Boost calls it) BEFORE I have read all the information on the socket. In these circumstances, I can continue reading information from the socket after receiving the close notification (or EOF). So, I'm starting to believe that hMailServer does indeed have a bug.

RvdHout:
@martinknafve 

I think i might have found the culprit causing this error, 

TCPConnection::AsyncRead uses transfer_at_least(1), which expect at least 1 byte to be transferred but if no more bytes are transferred as there no more bytes in the (file)stream, this results in the mentioned boost::asio::error::eof (boost error code 2, which is not a winsock error as reported by HMS)

> "DEBUG" 3424 "2015-05-15 16:22:21.717" "Parsing response from SpamAssassin. Session 30202"
> "DEBUG" 3424 "2015-05-15 16:22:21.811" "The read operation failed. **Bytes transferred: 0** Remote IP: 127.0.0.1, Session: 30202, Code: 2, Message: End of file"
> "DEBUG" 3424 "2015-05-15 16:22:21.811" "Ending session 30202"

RvdHout:
@martinknafve 
Some workarounds, that might fix this winsock error 2 issue with SA
https://www.hmailserver.com/forum/viewtopic.php?f=7&t=39288&p=241516#p241516

RvdHout:
@martinknafve 

Same here, https://github.com/hmailserver/hmailserver/issues/195

There you claim it is not a bug, but it is....  taken that example

`
"DEBUG"	3560	"2017-01-17 11:54:01.446"	"Creating session 204"
"TCPIP"	3560	"2017-01-17 11:54:01.446"	"TCP - xxx.xxx.xxx.xxx connected to xxx.xxx.xxx.xxx:110."
"DEBUG"	3560	"2017-01-17 11:54:01.446"	"TCP connection started for session 197"
[...]

PR #440, opened 2022-08-17
HM5157. Hmailserver winsock 2. There was a communication error with S…

RvdHout:
…pamAssassin

https://github.com/hmailserver/hmailserver/issues/167

### DIFF
--- a/hmailserver/source/Server/Common/AntiSpam/SpamTestSpamAssassin.cpp
+++ b/hmailserver/source/Server/Common/AntiSpam/SpamTestSpamAssassin.cpp
@@ -92,7 +92,7 @@ namespace HM
       else
       {
          String message = "The IP address for SpamAssassin could not be resolved. Aborting tests.";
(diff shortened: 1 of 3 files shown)


RvdHout:
FYI, what this does:

Ignore EOF (errorcode = 2) errors when 'if (receive_binary_)' (which is basically SA)
In all other cases do not ignore EOF errors

RvdHout:
Not sure if this is completely safe to do,

RvdHout:
@martinknafve 

So after running this for a month continuously I think it safe to conclude this proposed fix totally eliminates the "The WinSock error code is 2" with SpamAssassin without causing other issues.

RvdHout:
I updated this pull request as there seemed to be a bug in the previous approach that could result in hmailserver.exe peaking CPU usage when a EOF occurred during a SMTP command. With the new approach i distinguish between SMTP/IMAP/POP and the connection to SpamAssassin and only ignore EOF errors when it is SpamAssassin

PR #444, opened 2022-09-13
Automatically append ESMTP to Welcome-message if user has not specified it

RvdHout:
SMTPConnection::SendBanner_() fixes, check if ends with " ESMTP" and if not append it


cblaze22:
Could you do a pull to add new rule to bind to a local host name like the bind to ip does currently?

martinknafve (approved):

PR #447, opened 2022-10-18
Better singleton and boost::regex

xth:


martinknafve:
I agree this implementation looks better, but I don't think there's enough benefits with merging. If the old functionality _works_, I will keep it as is considering no major changes are made to hMailServer any more.

PR #457, opened 2023-03-29
5.7.0-SASL-Cleanup

RvdHout:
Remove (unused) leftovers from https://github.com/hmailserver/hmailserver/pull/82

I did not remove setting column ImapAuthAllowPlainText from SQL (yet) Do you want me to?


martinknafve:
Yes, the SQL database row should be removed if it's no longer needed.

Issue #462, opened 2023-05-23
MX record not resolved correctly , when initial hostname is referring to a CNAME


blackbaro:
Let's say we have this configuration

CNAME external.bankup.be =>  to application.bankup.be.
A application.bankup.be => 10.0.0.0  (private ip as testscenario)
MX application.bankup.be => 10 mail.bankup.be.
A mail.bankup.be. => 10.0.0.1 (private ip as testscenario)

the MX resolution of test@external.bankup.be should resolve to (private ip as testscenario) . The latest version of hmailserver v2.9 is resolving 10.0.0.0 instead. This used to work correctly in previous versions of hmailserver

Version 5.7.0-B2631
![image](https://github.com/hmailserver/hmailserver/assets/67577536/5f0a3885-8107-49b2-9bcd-9dc8204f4ab5)

Version 5.6.8-B2574
![image](https://github.com/hmailserver/hmailserver/assets/67577536/799548f8-8f8f-496d-b45d-276711d33e09)

RvdHout:
A MX record may not be a CNAME, so actually 5.7 is right and < 5.7 not

https://www.rfc-editor.org/rfc/rfc2181.txt

blackbaro:
> A MX record may **not** be a CNAME, so actually 5.7 is right and < 5.7 not https://www.rfc-editor.org/rfc/rfc2181.txt
> 
> So simply point mail.bankup.be to a A-Record

mail.bankup.be is an A record. or did you mean external.bankup.be cannot be a CNAME but must be an A & MX itself? then this would mean SPF records would have to be created as well for this subdomain.

external.bankup.be(could be other rootdomains as well) CNAME to  => application.bankup.be  MX to => mail.bankup.be A to => 10.0.0.1
test@external.bankup.be should refer to the MX record of application.bankup.be instead of the A record of application.bankup.be

note, we don't have control over external.bankup.be. Currently the same domain, but in production whitelabels are using this CNAME approach. to delegate their subdomain over to application.bankup.be.

blackbaro:
Just like here => 

https://www.nslookup.io/domains/external.bankup.be/dns-records/mx/
![image](https://github.com/hmailserver/hmailserver/assets/37797348/397dc005-821f-4577-8687-b6d56c5c9496)

RvdHout:
Again...

>    The domain name used as the value of a NS resource record, or part of
>    the value of a MX resource record must not be an alias.  Not only is
[...]

PR #463, opened 2023-05-25
MX record not resolved correctly , when initial hostname is referring to a CNAME

blackbaro:
https://github.com/hmailserver/hmailserver/issues/462

PR #464, opened 2023-05-29
DNS resolver fixes

RvdHout:
- Fix CNAME (domain) record, see: https://github.com/hmailserver/hmailserver/issues/462 
- Fix IP Address as MX, see: https://github.com/hmailserver/hmailserver/issues/462#issuecomment-1563173781 
- Enhancement Null MX

@martinknafve 

I created 3 (temporary) test in RegressionTests -> Infrastructure -> DnsResolution, which all pass

```
      [Test]
      public void NullMXRecordShouldNotResolve()
      {
         // If a MX record contains a Null MX no mail server should be returned
         var actualServer = _utilities.GetMailServer("test@allelassers.nl");

         Assert.AreEqual("", actualServer);
      }

      [Test]
      public void IPAddressAsMXRecordShouldResolve()
      {
         // Okay, this is an invalid MX record. The MX record should always contain 
         // a host name but in this case it appears an IP address. We'll be kind to
         // the domain owner and still deliver the email to him.
         var actualServer = _utilities.GetMailServer("test@ruud.allelassers.nl");

         Assert.AreEqual("89.20.83.133", actualServer);
      }

      [Test]
      public void CnameDomainRecordsShouldResolve()
      {
         // If a Domain name resolves to a CNAME record, the CNAME record should be followed.
         var actualServer = _utilities.GetMailServer("test@external.bankup.be");

         Assert.AreEqual("10.0.0.1", actualServer);
      }
```

FYI: One test i am unable to complete successfully, eg: InfinitelyRecursiveRecordsShouldNotResolve

### DIFF
--- a/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp
+++ b/hmailserver/source/Server/Common/TCPIP/DNSResolver.cpp
@@ -168,8 +168,28 @@ namespace HM
    }
 
    bool
-   DNSResolver::GetEmailServers(const String &sDomainName, std::vector<HostNameAndIpAddress> &saFoundNames )
+   DNSResolver::GetEmailServers(const String &sDomainName, std::vector<HostNameAndIpAddress> &saFoundNames)
    {
+      return GetEmailServersRecursive_(sDomainName, saFoundNames, 0);
+   }
+
+   bool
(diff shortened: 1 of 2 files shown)


martinknafve:
Is this PR supposed to be merged _instead of_ #463? (I think so, but I'm not sure)

RvdHout:
✅

PR #512, opened 2026-04-12
DKIM-sign all email sent from domain

martinknafve:


hmailserver/source/Server/Common/AntiSpam/DKIM/DKIM.cpp
```diff
@@ -631,20 +639,6 @@ namespace HM
          }
       }
       
-      AnsiString allowedHashes = dnsKeyParams.GetValue("h");
```
martinknafve, line 634: This was redundant. Same validation was performed twice.

PR #513, opened 2026-04-17
Fix: Emails stored without trailing newline

martinknafve:
The fix for #500 introduced an issue where email messages were sometimes missing a trailing newline on disk. This later lead to issues parsing the content.

The code has been updated so that the trailing newline is added, and the parsing has been made more robust.

### DIFF
--- a/hmailserver/source/Server/Common/Mime/Mime.cpp
+++ b/hmailserver/source/Server/Common/Mime/Mime.cpp
@@ -58,20 +58,32 @@ namespace HM
    }
 
    // search for string2 in string1 (strstr)
-   static const char* FindString(const char* pszStr1, const char* pszStr2, const char* pszEnd)
+   const char* FindString(const char* haystack,
+      const char* needle,
+      const char* haystackEnd) // one past last valid char
    {
-      pszEnd -= ::strlen(pszStr2);
-      const char *s1, *s2;
-      while (pszStr1 <= pszEnd)
+      if (haystack == NULL || needle == NULL || haystackEnd == NULL)
+         return NULL;
+
+      const size_t needleLength = std::strlen(needle);
+      if (needleLength == 0)
+         return haystack;
+
+      if (haystackEnd <= haystack)
+         return NULL;
+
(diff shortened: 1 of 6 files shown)


hmailserver/source/Server/Common/Mime/Mime.cpp
```diff
@@ -58,20 +58,32 @@ namespace HM
    }
 
    // search for string2 in string1 (strstr)
-   static const char* FindString(const char* pszStr1, const char* pszStr2, const char* pszEnd)
+   const char* FindString(const char* haystack,
```
martinknafve, line 61: I initially thought that this function was the cause, so I made the validation in it more robust. I no longer think it was the cause, but decided to keep the added validation.

hmailserver/source/Server/Common/Mime/Mime.cpp
```diff
@@ -1634,19 +1648,44 @@ namespace HM
...
+         const char* pszAfterBoundary = pszBound1 + strBoundary.size();
+
+         // Need at least 2 bytes available to distinguish a closing boundary
+         // ("--") from a normal part boundary ("\r\n").
+         if (pszAfterBoundary + 2 > pszEnd)
+            break;
+
+         if (pszAfterBoundary[0] == '-' && pszAfterBoundary[1] == '-')
          {
-            // Closing boundary: record end right after "--boundary--" (before any trailing CRLF)
```
martinknafve, line 1640: This code sometimes did not save a newline at the end of the file. When the file was later read, this caused an error since it searched for "\r\n" (a few lines up).

I changed the logic so that a trailing \r\n is added, and also tried making the parsing more robust.

hmailserver/source/Server/Common/Mime/Mime.h
```diff
@@ -35,6 +35,8 @@ using namespace std;
 // RFC 822 - Standard For The Format of ARPA Internet Text Message
 namespace HM
 {
+   const char* FindString(const char* pszStr1, const char* pszStr2, const char* pszEnd);
```
martinknafve, line 38: Included this in header so that it can be called by test functions.

PR #514, opened 2026-04-18
Use empty envelope address where possible

martinknafve:


RvdHout:
I should have left a commit comment, but somehow i am not able to see it (myself)

image although it shows up above

martinknafve:
> I should have left a commit comment, but somehow i am not able to see it (myself)
> 

It's easier to add comments under "Files changed" instead.

I have:

* Removed calls to `SetReturnPath`. As you say, that should be set upon final delivery into an account.
* Removed calls to `pMsg->SetFromAddress(_T(""))`. The envelope from address is empty by default for new messages, so there's no point in setting it.

PR #523, opened 2026-05-01
Message SpamFlag, optionally disable reply/forward when spam marked

RvdHout:
- Fix some spam scoring inconsistencies, for example the spam-counter was not updated in POP3ClientConnection.cpp when a message scored above the SpamDeleteThreshold
- Make sure both the SpamMarkThreshold and SpamDeleteThreshold are > 0 in both POP3ClientConnection.cpp as in SMTPConnection.cpp‎ (this was not in POP3ClientConnection.cpp before)
- By correcting the above, if either the SpamMarkThreshold and/or SpamDeleteThreshold are set to 0, their functionality is disabled

Add a internal (persistent) FlagSpam message flag as the options to mark spam below are optional and can be disabled
- Add X-hMailServer-Spam
- Add X-hMailServer-Reason
- Add to message subject

Use the internal FlagSpam message flag to (optionally) determine if a message should be forwarded, replied to if defined by either Rule or with a Account Forward/OOF when a message is considered/marked as being spam

WebAdmin, GUI changes and tests included

-----------------------------------

In RuleApplier.cpp the include MailerDaemonAddressDeterminer.h can be removed, there was also a orphaned std::shared_ptr<Account> emptyAccount; that wasn't used.
In SMTPVacationMessageCreator.cpp the SetReturnPath action can also be removed, as part of https://github.com/hmailserver/hmailserver/pull/514

### DIFF
--- a/hmailserver/source/DBScripts/CreateTablesMSSQL.sql
+++ b/hmailserver/source/DBScripts/CreateTablesMSSQL.sql
@@ -190,7 +190,9 @@ create table hm_accounts (
 	accountvacationexpires tinyint not null,
 	accountvacationexpiredate datetime not null,
 	accountpersonfirstname nvarchar(60) not null,
-	accountpersonlastname nvarchar(60) not null
+	accountpersonlastname nvarchar(60) not null,
+	accountvacationabortspamflagged tinyint not null,
+	accountforwardabortspamflagged tinyint not null
 ) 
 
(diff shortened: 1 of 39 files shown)


martinknafve (commented):
For future PRs, please keep the scope smaller (to 1 item) instead of multiple. Will make reviewing easier.

PR #527, opened 2026-05-14
Mime: Use registered field coder (+cleanup)

RvdHout:
cleanup / remove unused
fix char in while loop, start with first char, do pszData++; after

### DIFF
--- a/hmailserver/source/Server/Common/Mime/MimeCode.cpp
+++ b/hmailserver/source/Server/Common/Mime/MimeCode.cpp
@@ -158,7 +158,7 @@ namespace HM
 	   for (auto it=field_coders_.begin(); it!=field_coders_.end(); it++)
 	   {
 		   ASSERT((*it).first != NULL);
-		   if (::_stricmp(pszFieldName, (*it).first))
+		   if (!::_stricmp(pszFieldName, (*it).first))
 		   {
 			   FIELD_CODER_FACTORY pfnCreateObject = (*it).second;
 			   ASSERT(pfnCreateObject != NULL);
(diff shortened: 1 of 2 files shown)


hmailserver/source/Server/Common/Mime/MimeCode.cpp
```diff
@@ -778,7 +777,7 @@ namespace HM
 
 			   while (pszData < pszEnd)
 			   {
-				   char ch = *pszData++;
+               char ch = *pszData;
```
martinknafve, line 780: Was this changed only for readability, or? The PR description mentions "fix", but it looks like the behavior here is not changed.

RvdHout: doesn't  char ch = *pszData++;  select the 2nd char? and char ch = *pszData; the first char?

martinknafve: No, it will select the first character. It will:

1. Take the character currently pointed to by pszData
2. Store it in ch
3. Advance pszData to point to the next character

The behavior you describe (take second character) would happen if the code instead was `*(++pszData)`.

The current code uses postfix operator, which increments the value _after_ the expression is evaluated. Random source:
https://faculty.cs.niu.edu/~mcmahon/CS241/Notes/increment_and_decrement.html

I don't mind the change, but I wanted to clarify that it won't change the behavior.

RvdHout:
There are some (more) inconsistencies in MimeCode.cpp, find all instances where `while (pbData < pbEnd)` is called...
Sometimes (unsigned) `char ch = *pbData+`+; is used, and in some other cases (unsigned)` char ch = *pbData;` 

Don't know if it needs to be changed now you explained, but it somewhat confusing

Few examples

(https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/Mime/MimeCode.cpp#L222-L224)

https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/Mime/MimeCode.cpp#L264-L266

https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/Mime/MimeCode.cpp#L331-L333

https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/Mime/MimeCode.cpp#L367-L369

https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/Common/Mime/MimeCode.cpp#L480-L482

martinknafve:
I don't think we need to change code which is working already. Of course, it's not optimal that two different styles are used, but since they have the same effect it should not matter (assuming it's at least incremented every time).

PR #537, opened 2026-08-14
Update PersistentDomain.cpp

RvdHout:
Domain size calculation used the wrong column on MS SQL, SQL CE and PostgreSQL

### DIFF
--- a/hmailserver/source/Server/Common/Persistence/PersistentDomain.cpp
+++ b/hmailserver/source/Server/Common/Persistence/PersistentDomain.cpp
@@ -298,7 +298,7 @@ namespace HM
       {
          sSQL = "select sum(messagesize) as size from hm_messages "
                 "where messageaccountid in "
-                "(select accountdomainid from hm_accounts where accountdomainid = @DOMAINID) ";
+                "(select accountid from hm_accounts where accountdomainid = @DOMAINID)";
       }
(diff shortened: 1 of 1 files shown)


RvdHout:
Btw, why exactly do you differentiate between MS SQL, SQL CE and PostgreSQL and MySQL/MariaDB? 
Is that a compatibly issue with older MySQL versions? As with at least MySQL 5.x and MariaDB you shouldn't need to use the inner join, is it?

```
      if (DBType == DatabaseSettings::TypeMSSQLServer || DBType == DatabaseSettings::TypeMSSQLCompactEdition || DBType == DatabaseSettings::TypePGServer)
      {
         sSQL = "select sum(messagesize) as size from hm_messages "
                "where messageaccountid in "
                "(select accountid from hm_accounts where accountdomainid = @DOMAINID)";
      }
      else if (DBType == DatabaseSettings::TypeMYSQLServer)
      {
         sSQL = "select sum(messagesize) as size from hm_messages "
                "inner join hm_accounts on "
                "(hm_messages.messageaccountid = hm_accounts.accountid and hm_accounts.accountdomainid = @DOMAINID)";

      }  
```

martinknafve:
It's just a leftover from very old MySQL version support. I removed it here:
https://github.com/hmailserver/hmailserver/pull/541

Issue #536, opened 2026-08-12
Auto added Message-ID


RvdHout:
@martinknafve 

In [SMTPMessageHeaderCreator.cpp ](https://github.com/hmailserver/hmailserver/blob/master/hmailserver/source/Server/SMTP/SMTPMessageHeaderCreator.cpp#L72-L78) a Message-ID header is automatically inserted if there was none, is this a just convenience thingy?

Was just wondering, as with that addition SA rules that check for the existence of the Message-ID header will always pass (like: MISSING_MID and NO_HEADERS_MESSAGE), even if the original message had none

RvdHout:
I could always write a custom rule to trigger this, like:

```
describe    T_LOCAL_MSGID Message is using auto-inserted Message-Id
header		__LOCAL_MSGID Message-ID =~ /(?:(?:[0-9a-fA-F]){8}-(?:[0-9a-fA-F]){4}-(?:[0-9a-fA-F]){4}-(?:[0-9a-fA-F]){4}-(?:[0-9a-fA-F]){12})\@mail\.mydomain\.nl/i
meta        T_LOCAL_MSGID ( !ALL_TRUSTED && __HAS_MESSAGE_ID &&  __LOCAL_MSGID )
# score       T_LOCAL_MSGID 0.552 0.140 1.199 0.497 # n=1
```
Score is based on normal SA rule for missing Message-ID (eg: MISSING_MID), all tough currently commented as this rule is in test phase

RvdHout:
Also referenced here: https://github.com/hmailserver/hmailserver/pull/535

martinknafve:
(No need to tag me, I see issues added)

Frankly, I don't remember. It has always had that behavior. Created PR to change it: https://github.com/hmailserver/hmailserver/pull/542/

PR #542, opened 2026-08-15
Only add missing Message-ID for authenticated senders, #536

martinknafve:
Adding a Message-ID when the submitted message lacks one is a message
submission task (RFC 6409). hMailServer did it for every message received
over SMTP, including mail relayed to us by other servers.

The header is prepended during the DATA flush, before the post-transmission
spam tests run, so SpamAssassin was always handed a message that had a
Message-ID. Rules that check for a missing one (MISSING_MID,
NO_HEADERS_MESSAGE) could therefore never trigger, and the same applied to
rules and scripts inspecting headers further down the line.

Restrict the insertion to authenticated senders, where we act as the
submission server. Relayed messages now only get trace fields, as they
should.

Co-Authored-By: Claude Opus 5

### DIFF
--- a/hmailserver/source/Server/SMTP/SMTPMessageHeaderCreator.cpp
+++ b/hmailserver/source/Server/SMTP/SMTPMessageHeaderCreator.cpp
@@ -69,8 +69,13 @@ namespace HM
 
       String sComputerName = Utilities::ComputerName();
 
-      // Add Message-ID header if it does not exist.
-      if (!original_headers_->FieldExists("Message-ID"))
+      // Add Message-ID header if it does not exist. Adding a missing Message-ID is a
+      // message submission task (RFC 6409), so it's only done for authenticated senders.
+      // When another server relays a message to us we act as a relay rather than as a
+      // submission server, and should only add trace fields. Inserting a Message-ID in
+      // that case would hide from spam filters and rules further down the line that the
+      // original message didn't have one.
+      if (is_authenticated_ && !original_headers_->FieldExists("Message-ID"))
       {
          String sTemp;
          sTemp.Format(_T("Message-ID: %s\r\n"), Utilities::GenerateMessageID().c_str());
--- a/hmailserver/test/RegressionTests/SMTP/Basics.cs
+++ b/hmailserver/test/RegressionTests/SMTP/Basics.cs
@@ -261,6 +261,39 @@ namespace RegressionTests.SMTP
       }
 
       [Test]
+      [Category("SMTP")]
+      [Description("Issue 536: A missing Message-ID should be added for authenticated senders, since we act as a submission server for them.")]
+      public void TestMessageIDAddedForAuthenticatedSender()
+      {
+         SingletonProvider<TestSetup>.Instance.AddAccount(_domain, "test@example.test", "test");
+
+         // The simulator does not add a Message-ID of its own, so any Message-ID in the
+         // delivered message was added by the server.
+         string errorMessage;
+         var simulator = new SmtpClientSimulator();
+         simulator.Send(false, "test@example.test", "test", "test@example.test", "test@example.test",
+                        "Test subject", "Test body", out errorMessage);
+
+         var test = Pop3ClientSimulator.AssertGetFirstMessageText("test@example.test", "test");
+
+         Assert.IsTrue(test.Contains("Message-ID: <"));
+      }
+
+      [Test]
+      [Category("SMTP")]
+      [Description("Issue 536: A missing Message-ID should not be added to relayed messages, since that hides from spam filters that the original message had none.")]
+      public void TestMessageIDNotAddedForUnauthenticatedSender()
+      {
+         SingletonProvider<TestSetup>.Instance.AddAccount(_domain, "test@example.test", "test");
+
+         SmtpClientSimulator.StaticSend("someone@example.com", "test@example.test", "Test subject", "Test body");
+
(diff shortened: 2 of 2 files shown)


RvdHout:
I just checked my SA logs, got many MISSING_MID entries now but all for mail deliberately sent unauthenticated (like forms on websites hosted on the same server) so it's get passed thru SA (to catch spammers)
Critical mails are always sent authenticated, it just the (contact)forms on websites that are sent unauthenticated.

```
Sun Aug 16 22:06:15 2026 [-5608] info: spamd: clean message (-3.0/4.0) for (unknown):0 in 1.7 seconds, 3465 bytes.
Sun Aug 16 22:06:15 2026 [-5608] info: spamd: result: . -3 - ALL_TRUSTED,BAYES_00,HTML_MESSAGE,KAM_DMARC_STATUS,MIME_HTML_ONLY,MISSING_MID,TXREP scantime=1.7,size=3465,user=(unknown),uid=0,required_score=4.0,rhost=PCNAME,raddr=127.0.0.1,rport=2557,mid=(unknown),bayes=0.000000,autolearn=no autolearn_force=no,shortcircuit=no
Sun Aug 16 22:06:15 2026 [-5608] info: spamd: connection from PCNAME [127.0.0.1]:2566 to port 783, fd 6
Sun Aug 16 22:06:15 2026 [-5608] info: spamd: processing message (unknown) for (unknown):0
Sun Aug 16 22:06:17 2026 [-5608] info: spamd: clean message (-3.8/4.0) for (unknown):0 in 1.7 seconds, 3445 bytes.
Sun Aug 16 22:06:17 2026 [-5608] info: spamd: result: . -3 - ALL_TRUSTED,BAYES_00,HTML_MESSAGE,KAM_DMARC_STATUS,MIME_HTML_ONLY,MISSING_MID,TXREP scantime=1.7,size=3445,user=(unknown),uid=0,required_score=4.0,rhost=PCNAME,raddr=127.0.0.1,rport=2566,mid=(unknown),bayes=0.000117,autolearn=no autolearn_force=no,shortcircuit=no
Sun Aug 16 22:27:36 2026 [-5608] info: spamd: connection from PCNAME [127.0.0.1]:2940 to port 783, fd 6
```

Now i can update the website forms code to include a Message-ID by itself if it is sent unauthenticated as System.Net.Mail doesn't seem to do that automatically, or would it be a better and more efficient idea to also add the Message-ID when it is a authenticated and a local domain/localhost, using something var like IsLocalDomain and/or send thru LocalIPAddresses::Instance()->IsWithinLoopbackRange(iIPAdress))?

RvdHout:
Here are my finding on components i tend to use for sending all kind of mails/notifications on the server

System.Net.Mail (.NET) -> No Message-ID
hMailServer.Message -> No Message-ID
CDO.Message (VBScript) -> Writes a Message-ID itself
Send-MailMessage (PowerShell) -> No Message-ID

That is only, if send unauthenticated of course 

@martinknafve, what do you think? This change could have huge impact (mail ending up marked as Spam) for people that sent mails/notifications unauthenticated, eg: trough localhost

Why did you decide to do that only for authenticated users, RFC 6409 doesn't really require that, or did i misread that?

**Gemini's answer:**
> No. [RFC 6409](https://datatracker.ietf.org/doc/html/rfc6409) does not restrict adding a Message-ID only to authenticated users. Instead, it defines the behavior of a Message Submission Agent (MSA) that receives mail from users.The RFC specifies that an MSA is permitted (and in some cases expected) to add a missing Message-ID to ensure the message is wire-valid, regardless of whether the user is authenticated.

RvdHout:
Could this be a viable solution? I think it will minimise impact for users but it still won't add Message-ID's for mail received from remote IP addresses

```
      IPAddress address;
      bool bLocalIPAddress = (address.TryParse(remote_ip_address_) && LocalIPAddresses::Instance()->IsWithinLoopbackRange(address));

      if ((is_authenticated_ || bLocalIPAddress ) && !original_headers_->FieldExists("Message-ID"))
      {
         String sTemp;
         sTemp.Format(_T("Message-ID: %s\r\n"), Utilities::GenerateMessageID().c_str());
         new_header_lines += sTemp;
      }
```

Your Regression tests need a rework though

RvdHout:
Or like this, but this might be overkill

```
      IPAddress address;
      bool bLocalIPAddress = (address.TryParse(remote_ip_address_) && LocalIPAddresses::Instance()->IsWithinLoopbackRange(address));

      std::shared_ptr<DomainAliases> pDA = ObjectCache::Instance()->GetDomainAliases();
      String primaryAddress = pDA->ApplyAliasesOnAddress(message_->GetFromAddress());
      String primaryDomain = StringParser::ExtractDomain(primaryAddress);
      std::shared_ptr<const Domain> pDomain = CacheContainer::Instance()->GetDomain(primaryDomain);
      bool bDomainIsLocal = pDomain != nullptr;

      if ((is_authenticated_ || bLocalIPAddress && bDomainIsLocal) && !original_headers_->FieldExists("Message-ID"))
      {
         String sTemp;
         sTemp.Format(_T("Message-ID: %s\r\n"), Utilities::GenerateMessageID().c_str());
         new_header_lines += sTemp;
      }
```

martinknafve:
Agreed, this needs fixing — RFC 6409 keys on submission, not authentication. Proposed fix in #552: instead of hard-coding loopback, it reuses the submission/relay decision already made at RCPT time, so unauthenticated local senders from an IP range that doesn't require auth (localhost by default) get a Message-ID again, while relayed mail still doesn't.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #553, opened 2026-08-17
Flush log files when they are kept open, #549

### LINKED ISSUE #549: LogFile Flush when no activity
Sometimes, often in the evening/night when activity is low, it can take quite some time for the logfile to show up the last received command or returned results, can you ask Claude to flush the Log (files) on a set interval when no new activity? Does Claude produce better code then Copilot?

I just tested earlier posted code, but apparently asked Copilot the wrong question and it only flushed the LiveLog but not the LogFile itself

martinknafve:
Log data is written through a CRT-buffered `FILE*`. With "Keep files open" enabled the file is never closed after a write, so on an idle server the last lines sit in the buffer instead of appearing in the log. Adds `File::Flush` and flushes after each write when the file is kept open.

Also fixes a shadowed `fileExists` local in `Logger::GetCurrentLogFile_` that made it always false — which reopened the file on every write despite "Keep files open", and wrote a BOM before every line of the backup and event logs.

Fixes #549

---
_Generated by [Claude Code](https://claude.ai/code/session_01TyYuCCqSoNGMDUaYL1Efwo)_

### DIFF
--- a/hmailserver/source/Server/Common/Application/Logger.cpp
+++ b/hmailserver/source/Server/Common/Application/Logger.cpp
@@ -326,7 +326,7 @@ namespace HM
 
       try
       {
-         bool fileExists = FileUtilities::Exists(fileName);
+         fileExists = FileUtilities::Exists(fileName);
       }
       catch (boost::system::system_error&)
       {
@@ -409,7 +409,13 @@ namespace HM
          file->Write(sAnsiString);
       }
 
-      if (!keepFileOpen)
+      if (keepFileOpen)
+      {
+         // The file is buffered by the CRT and won't be closed after this write. Flush it, so that
+         // the last lines don't remain invisible in the log file while the server is idle.
+         file->Flush();
+      }
+      else
          file->Close();
    }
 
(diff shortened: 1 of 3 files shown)


RvdHout:
This a replacement for the flushing based on fixed interval?

martinknafve:
Yes. If you look at the code that was changed, there was a bug that caused the file to be closed/opened on every write:

```
      bool fileExists = false;

      try
      {
         bool fileExists = FileUtilities::Exists(fileName);
      }
      catch (boost::system::system_error&)
      {
         // If the log is not accessible for some reason, such as error code 5 - access is denied, 
         // then we can't open the log file. This will result in nothing being logged.
         return nullptr;
      }

      if (file->IsOpen())
      {
         if (file->GetName() != fileName)
            file->Close();
         else if (!fileExists)
            file->Close();
      }
```

Note that the first line defines `fileExists`. Then when we check if the file exists, we create a _new_ bool and set the value to it. The end result is that fileExists will always be false in the following lines, which results in the file being closed and re-opened for every write.

This PR changes this so that the file is _actually_ kept open, and instead flush the content. It should solve your issue and also give better performance.

Issue #534, opened 2026-06-24
RFC 6154 - IMAP LIST Extension for Special-Use Mailboxes


RvdHout:
Hi @martinknafve, 
Have you ever considered supporting the SPECIAL-USE IMAP extension?

_As you may know I am Dutch, and every time i add a IMAP account (hosted within hMailServer) in Outlook the default folders are not what one might assume, eg: INBOX, Sent, Trash, Drafts, Junk , but for example (a Dutch version of) Outlook creates folders like:  INBOX, Verzonden items, Verwijderde items, Concepten, Ongewenste e-mail, the IMAP SPECIAL-USE extension is supposed to fix this, summary:_

IMAP SPECIAL-USE is an extension (standardized in [RFC 6154](https://datatracker.ietf.org/doc/html/rfc6154)) that allows an email server to automatically tag specific folders with attributes defining their function (like Drafts, Sent, or Trash). It prevents email clients from creating duplicate folders by hinting at an intended purpose regardless of the folder's name.

### Why SPECIAL-USE is Important

**Folder Unification:** Instead of different email apps creating "Sent Items," "Sent Messages," or localized versions, the server tags one folder with the \Sent attribute so all clients sync to the exact same location.
**Automated Configuration:** Compliant email clients read the server's tags during the initial setup and map the correct folders automatically.

### Common SPECIAL-USE Attributes

Servers and email providers like Gmail (IMAP Extensions | Gmail) utilize the following standard and extended attributes: 
\Drafts: Messages being composed but not yet sent. 
\Sent: Messages successfully sent by the user. 
\Trash: Deleted messages marked for removal. 
\Junk: Spam or unwanted messages.
\Archive: Mailboxes used to store old messages out of the user's way.
\Important: Flags specific important messages. 
\All: A virtual mailbox presenting all messages in the user's message store

martinknafve:
I think it could be added relatively easily, but would require database changes. Do you have database changes in your fork that are not in this repo? (Just wondering how messy it will be for you if I add new columns in this repo).

RvdHout:
No, currently on 5708 == same as the master repo

[...]

PR #555, opened 2026-08-17
Fix special-use IMAP folder deletion leaving orphaned DB row, #534

martinknafve:
PersistentIMAPFolder::DeleteObject() intentionally skips the SQL DELETE
for the Inbox and special-use folders (Sent, Drafts, Junk, Trash,
Archive), but always returned true either way. Every caller up the
chain treated that as full success:

- Collection<T,P>::DeleteItemByDBID() discarded P::DeleteObject()'s
  result, always erased the item from the in-memory collection, and
  always returned true.
- InterfaceIMAPFolder::Delete() (the COM method hMailAdmin calls) then
  always returned S_OK.
- hMailAdmin's delete-folder handler never checked the result and
  unconditionally removed the tree node.

So deleting a special-use folder in hMailAdmin removed it from the UI
and wiped its messages/permissions, but left an empty, still
special-use-flagged row in hm_imapfolders that reappeared on the next
folder tree reload.

Collection::DeleteItemByDBID now only erases from the in-memory list
and reports success when the underlying delete actually happened, and
InterfaceIMAPFolder::Delete() now surfaces the failure as a COM error
instead of swallowing it, so hMailAdmin (and any other COM caller)
sees an exception and leaves the folder in place.

### DIFF
--- a/hmailserver/source/Server/COM/InterfaceIMAPFolder.cpp
+++ b/hmailserver/source/Server/COM/InterfaceIMAPFolder.cpp
@@ -295,13 +295,13 @@ STDMETHODIMP InterfaceIMAPFolder::Delete()
       if (!parent_collection_)
       {
          if (!HM::PersistentIMAPFolder::DeleteObject(object_))
-            return COMError::GenerateError("The folder could not be deleted. The account Inbox and folders with a special-use flag (such as Sent, Drafts, Junk, Trash or Archive) cannot be deleted.");
+            return COMError::GenerateError("The folder could not be deleted. Please check the hMailServer error log for details.");
 
          return S_OK;
       }
 
       if (!parent_collection_->DeleteItemByDBID(object_->GetID()))
-         return COMError::GenerateError("The folder could not be deleted. The account Inbox and folders with a special-use flag (such as Sent, Drafts, Junk, Trash or Archive) cannot be deleted.");
+         return COMError::GenerateError("The folder could not be deleted. Please check the hMailServer error log for details.");
 
       return S_OK;
    }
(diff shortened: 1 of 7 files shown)


RvdHout:
Both test do not play nice... both fail

```
  Expected: <System.Runtime.InteropServices.COMException>
  But was:  null
```

Oh, nevermind, i think i know what might be the issue.... interop must be build as well, no?
[edit] no still the same

martinknafve:
The PRs created by Claude (branch name start with Claude) aren't really ready for use. It's code that Claude has added that I have not yet reviewed or tested.

I think the core issue is that I'm unsure about expected behavior. Should a user be able to delete special use folders? I think explicit delete of a specific folder via hmailadmin or Imap client should be allowed. Do you agree? 

Otherwise you might end up with multiple folders with same special use, like "Sent" and "Skickade" (sent in Swedish).

RvdHout:
Can you have multiple folders with the same special-use flags? Can == allowed by RFC

Maybe it should be able to delete a single folder via COM and/or client, but keep empty account via COM as is, eg: retain special-use  folders

martinknafve:
> Can you have multiple folders with the same special-use flags? Can == allowed by RFC

The RFC leaves this up to the implementor to decide. hMailServer blocks IMAP commands that try to create 2 with the same purpose. It's not blocked in the API at this point, was thinking of fixing that later.

```
   Servers MAY allow the creation of a special-use mailbox even if one
   so designated already exists.  This might have the effect of moving
   the special use from the old mailbox to the new one, or might create
   multiple mailboxes with the same special use.  Alternatively, servers
   MAY refuse the creation, considering the designation to be a
   conflict.
```


> Maybe it should be able to delete a single folder via COM and/or client, but keep empty account via COM as is, eg: retain special-use folders

Yes, that is what this branch now implements (after the most recent commit). All the regression tests pass and I have verified that you can delete special folders using hMailAdmin and IMAP DELETE command. I will run it on all database types before merging.

One thing to note is that hMailAdmin also allows you to delete INBOX. That was the case in 5.6 as well. This feels a bit odd to me, but not changing that now.

PR #568, opened 2026-08-25
Avoid endless read loop on peer-closed stream

hgy29:
My hMailServer used to enter endless CPU loops, ending up consuming all CPU on my server. I tracked it down to the TCP socket reading loop, which was constantly receiving 0 bytes while EOF condition was reported and yet reading was reissued.
The proposed fix addresses the issue for me. My understanding is that when EOF is reached and no more bytes are returned, then the socket has been closed by the peer and there is no point trying to fetch more bytes from it.

### DIFF
--- a/hmailserver/source/Server/Common/TCPIP/TCPConnection.cpp
+++ b/hmailserver/source/Server/Common/TCPIP/TCPConnection.cpp
@@ -493,8 +493,8 @@ namespace HM
    {
       UpdateAutoLogoutTimer();
 
-      // Ignore end of file or end of stream error when binary transfer, there may still be data in the receive buffer we can read.
-      if ((error && !receive_binary_) || (error && error != boost::asio::error::eof && receive_binary_))
+      // Ignore end of file or end of stream error when binary transfer and we received some bytes, there may still be data in the receive buffer we can read.
+      if ((error && !receive_binary_) || (error && ((error != boost::asio::error::eof) || (bytes_transferred == 0)) && receive_binary_))
       {
          if (connection_state_ != StateConnected)
          {


RvdHout:
I think i had that happen to me before, back then i fixed it using:

```
            try
            {
               if (pBuffer->GetSize() > 0)
                  ParseData(pBuffer);
            }
            catch (DisconnectedException&)
            {
               throw;
            }
            catch (...)
            {
               String message;
               message.Format(_T("An error occured while parsing data. Data size: %d"), pBuffer->GetSize());

               ReportError(ErrorManager::Medium, 5136, "TCPConnection::AsyncReadCompleted", message);

               throw;
            }
```

But after a while i started to question myself, mainly if it made any sense to do that in the try, catch
...then removed it and never ran into that same CPU exhausting behavior you now report

hgy29:
Not sure either if I am fixing it the best way. 
From what I understood, ParseData() is responsible for enqueuing a disconnection when it sees fit, which it probably won't do if no more bytes are received. Unless the data flow relies on some timeout, I assumed that a real EOF is an unexpected condition. It doesn't seem to break anything SMTP or IMAP related, can't tell for POP since I don't use it.
To give a bit more context, the issues arises sometimes when receiving an email through ASSP. ASSP opens a SMTP connection to hMailServer and starts relaying traffic, but can abruptly close the connection if it detects malformed content.

RvdHout:
Yes, i confirm it was a very rare occurrence for me as well... likely @martinknafve will have a look now and decide what the best fix is

RvdHout:
If i run it trough copilot, i get those recommendations

What changed and why
•	Replaced the long combined boolean expression with isEof and ignoreEofInBinary booleans to make the logic explicit and easier to audit.
•	Used early returns for the error handling path so the normal data-processing path is not nested inside a large else branch.
•	Behavior preserved: any error in non-binary mode is handled as before; in binary mode EOF is ignored only when bytes were received (so remaining buffer can be processed).
•	This clarifies intent and reduces risk of precedence/parentheses bugs.

```
   void 
   TCPConnection::AsyncReadCompleted(const boost::system::error_code& error, size_t bytes_transferred)
   {
      UpdateAutoLogoutTimer();

      // Ignore end of file or end of stream error when binary transfer and we received some bytes,
      // there may still be data in the receive buffer we can read.
      if (error)
      {
         const bool isEof = (error == boost::asio::error::eof);
         const bool ignoreEofInBinary = receive_binary_ && isEof && (bytes_transferred > 0);

         if (!ignoreEofInBinary)
         {
            if (connection_state_ != StateConnected)
            {
               // The read failed, but we've already started the disconnection. So we should not log the failure
               // or enqueue a new disconnect.
               return;
            }

            if (isEof)
            {
               // Ignore end of file or end of stream error
               return;
            }

            OnReadError(error.value());

            String message;
            message.Format(_T("The read operation failed. Bytes transferred: %d"), bytes_transferred);
            ReportDebugMessage(message, error);

            if (error == boost::asio::error::not_found)
            {
               // read buffer is full...
               OnExcessiveDataReceived();
            }

            EnqueueDisconnect();
            return;
         }
         // else: ignore EOF for binary case where bytes_transferred > 0 and continue processing buffer
      }

      if (receive_binary_)
      {
         std::shared_ptr<ByteBuffer> pBuffer = std::shared_ptr<ByteBuffer>(new ByteBuffer());
         pBuffer->Allocate(receive_buffer_.size());

         std::istream is(&receive_buffer_);
         is.read((char*)pBuffer->GetBuffer(), receive_buffer_.size());

         try
         {
            if (pBuffer->GetSize() > 0)
               ParseData(pBuffer);
         }
         catch (DisconnectedException&)
         {
            throw;
         }
         catch (...)
         {
            String message;
            message.Format(_T("An error occured while parsing data. Data size: %d"), pBuffer->GetSize());

            ReportError(ErrorManager::Medium, 5136, "TCPConnection::AsyncReadCompleted", message);

            throw;
         }
      }
      else
      {
         std::string s;
         std::istream is(&receive_buffer_);
         std::getline(is, s, '\r');

         // consume trailing \n on line.
         receive_buffer_.consume(1);

#ifdef _DEBUG
         String sDebugOutput;
         sDebugOutput.Format(_T("RECEIVED: %s\r\n"), String(s).c_str());
         OutputDebugString(sDebugOutput);
#endif

         try
         {
            ParseData(s);
         }
         catch (DisconnectedException&)
         {
            throw;
         }
         catch (...)
         {
            String message;
            message.Format(_T("An error occured while parsing data. Data length: %d, Data: %s."), s.size(), String(s).c_str());

            ReportError(ErrorManager::Medium, 5136, "TCPConnection::AsyncReadCompleted", message);

            throw;
         }
      }

      operation_queue_.Pop(IOOperation::BCTRead);
      ProcessOperationQueue_(0);
   }
```

martinknafve:
Thank you! The fix in this PR introduces a new issue where a message will not be delivered if the sender shut down sending directly after sending the message. RvdHouts fix did not have that issue, so I prefer that one. Since this PR would introduce another issue, I created a separate PR with his fix. The newly added regression test `PipelinedMessageFollowedByImmediateCloseShouldBeDelivered` triggers the issue that this PR caused.

As for Copilot recommendations, that new code is probably objectively better, but I prefer to make tiny incremental changes, so I'll leave that for now. Maybe when I have worked through more high-prio things, I can spend time on making the code itself more readable.

I'll close that if it's fine with you?

Here's the other PR:
https://github.com/hmailserver/hmailserver/pull/573

hgy29:
Yes it is fine as long as the issue is fixed.
Thank you!

PR #575, opened 2026-08-26
Document CreateDefaultSpecialUseFoldersEnabled IMAP option

martinknafve:
The IMAP folders doc claimed hMailServer never decides special-use
folder names/contents, which is now only true by default: the
CreateDefaultSpecialUseFoldersEnabled setting (Settings > Protocols >
IMAP > Advanced) makes hMailServer create Drafts/Sent/Trash/Junk with
RFC 6154 SPECIAL-USE attributes for new accounts. Documented the
exception and added the setting to the IMAP settings reference page.


RvdHout:
CreateDefaultSpecialUseFoldersEnabled is set to 1 (eg: enabled) by default, not? 

If not intended to be enabled by default the database initial and update scripts for dbversion 5709 are wrong

martinknafve:
You can ignore commits by Claude (and comments on PRs by Claude). They aren't reviewed, they are drafts that I'll look into later. 

But thanks, Claude corrected its mistake. I think, I'll check later in week. 😁

RvdHout:
Hey, if i can learn Claude a thing or two i am happy 😄

Issue #48, opened 2014-10-20
Error: Task-DeliveryTask


mseachrist:
I am running version 5.5.1 B2107.  I have been receiving a few errors but doesn't appear to be causing a major problem with the server.  The email server continues working fine.  I would just like to point out these errors to see what needs done.  I can send logs and minidumps at request.  

"Severity: 2 (High), Code: HM4208, Source: ExceptionHandler::Run, Description: An error occured while executing 'Task-DeliveryTask', Message: The client has been disconnected."

PR #605, opened 2026-09-02
Never leave the greeting state without a HELO/EHLO

martinknafve:
Follow-up to #604, which fixed the RSET half of this.

## The shape of the bug

`ResetCurrentMessage_` ended with an unconditional `current_state_ = HEADER`. `HELO`/`EHLO` are the only intended `INITIAL -> HEADER` transitions, and both guard their own assignment - so any other caller of `ResetCurrentMessage_` silently promoted an ungreeted session past the greeting. #604 fixed that for `RSET` by restoring the state in `ProtocolRSET_` afterwards.

Two things were left:

1. **The guard belongs in `ResetCurrentMessage_`.** Fixing it there covers all nine callers rather than one, and leaves no trap for the next caller. None of the others can be in `INITIAL` - they run off DATA completion, QUIT, `OnExcessiveDataReceived` or the destructor - so this is a no-op for them.

2. **The STARTTLS path had the same bug.** `OnHandshakeCompleted` cleared `helo_host_` and then called `ResetCurrentMessage_`, landing the session in `HEADER`. A client could send `MAIL FROM` without greeting the encrypted session at all, and the message was accepted with no HELO host: the HELO-based spam tests ran against an empty string, the `Received:` header fell back to the bare IP, and `OnHELO`/`OnEHLO` never fired for the session that actually carried the mail.

## What the RFCs say

RFC 3207 section 4.2 says the handshake resets the protocol "to the initial state (the state in SMTP after a server issues a 220 service ready greeting)" and that the server MUST discard the EHLO argument. RFC 5321 section 4.1.4 then puts a MUST on the client: "a session that will contain mail transactions MUST first be initialized by the use of the EHLO command."

So a client that skips the second EHLO and sends `MAIL FROM` is out of order - it is not exercising RFC 3207's `SHOULD send an EHLO command as the first command`. That SHOULD covers the commands which need no initialization, and those still work: `NOOP`, `HELP`, `QUIT`, `RSET`, `HELO` and `EHLO` are all dispatched ahead of the state switch in `InternalParseData`. Section 4.1.4 is explicit that `RSET` in particular should be answered normally rather than with a 503, which is what #604 already does.

The three places that decide whether a session has been greeted now quote the text they implement, so the behaviour does not have to be taken on trust.

## Behaviour change worth a release note

Requiring a greeting after the handshake will 503 a client that previously got through.

`AUTH` is the case to watch. It is advertised before TLS on a STARTTLS-**optional** port (`GetAuthIsEnabled_() && (IsSSLConnection() || GetConnectionSecurity() != CSSTARTTLSRequired)`), so a client can learn the mechanism from the plaintext EHLO and previously went straight to `AUTH` after the handshake. Such a client now has to send `EHLO` first, as it already must on a STARTTLS-**required** port, where `AUTH` is not advertised until after TLS.

hMailServer's own outbound client is unaffected - `SMTPClientConnection::OnHandshakeCompleted` already calls `ProtocolStateHELOEHLO_`. Two places in the test suite did skip the re-greeting and are corrected here; that RFC 3207 asks clients to send it anyway is why those edits are no-ops against the old server.

The commit that makes this change is separate from the rest so it can be reverted on its own.

## Tests

`SSL/StartTls/SmtpServerTests.cs`

- `AfterStartTlsMailFromShouldRequireNewEhlo`
- `AfterStartTlsRsetShouldNotAllowMailFrom`
- `AfterStartTlsMailFromShouldBeAcceptedOnceEhloHasBeenReissued`

`SMTP/Basics.cs` - replacing `TestPreventRSETByPassHELO`, which asserted against `503 Bad sequence of command` (missing the trailing s) and carried two `DisconnectInvalidClients`/`MaxNumberOfInvalidCommands` settings that had nothing to do with what it tested:

- `RsetBeforeGreetingShouldNotAllowMailFrom` - now also asserts the `RSET` itself is answered `250 OK`
- `RsetAfterGreetingShouldStillAllowMailFrom`
- `RsetShouldAbortTransactionInProgress`

The last two exist so that this cannot be "fixed" by breaking `RSET`.

The tests were pushed on their own first and confirmed red on exactly the two expected tests and nothing else (981 passed / 2 failed / 12 skipped), then green once the fix landed (983 passed / 0 failed / 12 skipped) in run [#48](https://github.com/hmailserver/hmailserver/actions/runs/33649960536).

---
_Generated by [Claude Code](https://claude.ai/code/session_014iDHFrsVSMhsrXmv4tEnMq)_

### DIFF
--- a/hmailserver/source/Server/SMTP/SMTPConnection.cpp
+++ b/hmailserver/source/Server/SMTP/SMTPConnection.cpp
@@ -150,11 +150,13 @@ namespace HM
                GetConnectionSecurity() == CSSTARTTLSRequired)
       {
          /*
-           Upon completion of the TLS handshake, the SMTP protocol is reset to
-           the initial state (the state in SMTP after a server issues a 220
-           service ready greeting). The server MUST discard any knowledge
-           obtained from the client, such as the argument to the EHLO command,
-           which was not obtained from the TLS negotiation itself.
+           RFC 3207, 4.2:
+
+             Upon completion of the TLS handshake, the SMTP protocol is reset to
+             the initial state (the state in SMTP after a server issues a 220
+             service ready greeting). The server MUST discard any knowledge
+             obtained from the client, such as the argument to the EHLO command,
+             which was not obtained from the TLS negotiation itself.
          */
 
          helo_host_.Empty();
@@ -162,8 +164,16 @@ namespace HM
          ResetLoginCredentials_();
          ResetCurrentMessage_();
 
-         // The session is back at the point right after the 220 greeting, so the
-         // client has to send a new EHLO before it can issue any other command.
+         /*
+           The initial state is the state before any greeting, so the client has to
+           greet us again before it can send mail. RFC 5321, 4.1.4:
+
+             A session that will contain mail transactions MUST first be
+             initialized by the use of the EHLO command.
+
+           The commands which need no initialization are unaffected, since they are
+           dispatched ahead of the state switch in InternalParseData.
+         */
          current_state_ = INITIAL;
 
          EnqueueRead();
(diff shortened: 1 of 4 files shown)


RvdHout:
I see there is/was more to it, the changes in this make sense i have to admit

cybercode3:
Good to know there was more to address beyond the original fix.

RvdHout:
@martinknafve , you not going to commit this? It seems there were scenarios left out that could make it possible to bypass HELO/EHLO, eg: Claude did i better job then me/us

martinknafve:
Hmm yes, just haven't had time to look at it until now.

Issue #608, opened 2026-09-03
PTR lookup fail for localhost (::1/127.0.01)


RvdHout:
i got a long outstanding issue i cannot wrap my head around, i'll try to explain: 
mail coming from localhost (::1/127.0.0.1) always fail PTR lookups, like:

> "TCPIP"	2660	"2026-09-03 12:25:53.301"	"TCP - ::1 connected to ::1:587."
> "SMTPD"	2660	6	"2026-09-03 12:25:53.340"	"::1"	"SENT: 220 Welcome to mail.mydomain.nl ESMTP"
> "SMTPD"	2660	6	"2026-09-03 12:25:53.344"	"::1"	"RECEIVED: HELO NUC8i3BEH"
> "SMTPD"	2660	6	"2026-09-03 12:25:53.385"	"::1"	"SENT: 250 Hello."
> "SMTPD"	9144	6	"2026-09-03 12:25:53.387"	"::1"	"RECEIVED: MAIL FROM: <postmaster@mydomain.nl>"
> "DEBUG"	9144	"2026-09-03 12:25:53.396"	"Spam test: SpamTestHeloHost, Score: 0"
> "DEBUG"	9144	"2026-09-03 12:25:53.397"	"Spam test: SpamTestPTR, Score: 1" **_<!-- notice the PTR check score_**
> "DEBUG"	9144	"2026-09-03 12:25:53.473"	"Spam test: SpamTestMXRecords, Score: 0"
> "DEBUG"	9144	"2026-09-03 12:25:53.475"	"Spam test: SpamTestSPF, Score: 0"
> "DEBUG"	9144	"2026-09-03 12:25:53.477"	"Total spam score: 1"
> "SMTPD"	9144	6	"2026-09-03 12:25:53.481"	"::1"	"SENT: 250 OK"
> "SMTPD"	7628	6	"2026-09-03 12:25:53.491"	"::1"	"RECEIVED: RCPT TO: <ruud@mydomain.nl>"
> "SMTPD"	7628	6	"2026-09-03 12:25:53.513"	"::1"	"SENT: 250 OK"
> "SMTPD"	9260	6	"2026-09-03 12:25:53.519"	"::1"	"RECEIVED: DATA"
> "SMTPD"	9260	6	"2026-09-03 12:25:53.585"	"::1"	"SENT: 354 OK, send."
[...]

PR #609, opened 2026-09-03
DNSResolverWinApi

RvdHout:
- Multiple calls to `free(pSrvList);` when `pSrvList != NULL`, on errors
- Use proper bitwise operators for fOptions

### DIFF
--- a/hmailserver/source/Server/Common/TCPIP/DNSResolverWinApi.cpp
+++ b/hmailserver/source/Server/Common/TCPIP/DNSResolverWinApi.cpp
@@ -72,16 +72,19 @@ namespace HM
 
       if (!IniFileSettings::Instance()->GetUseDNSCache())
       {
-         fOptions += DNS_QUERY_BYPASS_CACHE;
+         fOptions |= DNS_QUERY_BYPASS_CACHE;
       }
 
+      // Use a Custom DNS server IPv4 address if specified in the settings. 
+      // If the address is invalid, we will fallback to the system DNS servers.
       AnsiString sCustomDNS;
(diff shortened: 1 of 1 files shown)


RvdHout:
if you do not like this bit:

```
         // On some systems the in-addr.arpa or ip6.arpa addresses will have a trailing dot for local ip addresses in the PTR lookup. 
         // Remove it to ensure we can compare the query with the result.
         if (pDnsRecord->wType == DNS_TYPE_PTR) 
         {
            if (!name.empty() && name.back() == '.') {
               name.pop_back();
            }
         }
```

Other possible solution could be to [replace](https://github.com/RvdHout/hmailserver/blob/6ec78cbc060bc2a372953c0977b4e8d1d3c5f02a/hmailserver/source/Server/Common/TCPIP/DNSResolverWinApi.cpp#L157-L158)

```
         if (pDnsRecord->wType == resourceType && 
             query.Equals(name))
```

with: (.StartsWith(query) or .Contains(query)/.ContainsNoCase(query))

```
         if (pDnsRecord->wType == resourceType && 
             name.StartsWith(query))
```

as the **query** would hold: 
```
1.0.0.127.in-addr.arpa
1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa 
```
and the **name** on Windows Server 2025 would hold: 
```
1.0.0.127.in-addr.arpa.
1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa.
```

https://github.com/hmailserver/hmailserver/issues/608

PR #642, opened 2026-09-13
Widen the SPF result to the seven results of RFC 7208

martinknafve:
SPF::Result had three values, Neutral, Fail and Pass, and every result the SPF library reported which was not a pass or a fail was collapsed onto Neutral. It now carries all seven results of RFC 7208 section 2.6, and the Authentication-Results header knows the name of each of them.

Translating the library's result masks off the reason bits, which the library ORs onto its return value in three places and leaves to the caller to remove. Without that, "no SPF record" is not recognized as such.

Test narrows what it returns to Pass, Fail and Neutral, the results it has always returned, so no message is treated differently than before.


RvdHout:
Interesting, this would make it easier to include the Received-SPF header?

As this might be useful for SpamAssassin, eg: it scans incoming headers for existing Received-SPF metrics added by hms. If found, it reuses that result to avoid unnecessary DNS lookups.

My implementation in mod, https://github.com/RvdHout/hmailserver/blob/5.7.0-Mod/hmailserver/source/Server/SMTP/SPF/SPF.cpp

But this is sub optimal, as it checks SPF twice (once for the SpamCheckSPF result , once for the Received-SPF header)

martinknafve:
Yeah. This is part of a larger change to replace RMSPF library. The current implementation is not easily testable, and it uses low level memory operations and native Windows APIs (instead of the existing classes in hMailServer). So this PR is preparation work for a new implementation.

Next step is adding tests based on the official test suite for SPF, http://www.open-spf.org/svn/project/test-suite/rfc4408-tests-yml/.

RvdHout:
How would you suggest to re-implement the a `Received-SPF`, eg: https://datatracker.ietf.org/doc/html/rfc7208#section-9.1 header with the new SPF parser?
Unfortunately SA doesn't check the `Authentication-Results` header (yet) to skip doing SPF checks within SA, it sorely relies on the `Received-SPF` header for this, eg: https://spamassassin.apache.org/full/4.0.x/doc/Mail_SpamAssassin_Plugin_SPF.html

I know a SPF check is a relatively light check, but it seems a bit silly to do it again in SA when hMailServer has just done the check

> **ignore_received_spf_header (0|1) (default: 0)**
> By default, to avoid unnecessary DNS lookups, the plugin will try to use the SPF results found in any Received-SPF headers it finds in the message that could only have been added by an internal relay.
> 
> Set this option to 1 to ignore any Received-SPF headers present and to have the plugin perform the SPF check itself.
> 
> Note that unless the plugin finds an identity=helo, or some unsupported identity, it will assume that the result is a mfrom SPF check result. The only identities supported are mfrom, mailfrom and helo.
> 
> **use_newest_received_spf_header (0|1) (default: 0)**
> By default, when using Received-SPF headers, the plugin will attempt to use the oldest (bottom most) Received-SPF headers, that were added by internal relays, that it can parse results from since they are the most likely to be accurate. This is done so that if you have an incoming mail setup where one of your primary MXes doesn't know about a secondary MX (or your MXes don't know about some sort of forwarding relay that SA considers trusted+internal) but SA is aware of the actual domain boundary (internal_networks setting) SA will use the results that are most accurate.
> 
> Use this option to start with the newest (top most) Received-SPF headers, working downwards until results are successfully parsed.

